PR #1, opened 2018-04-27
rails version to 5.1.6

Sushmareddy-OSI:
Made changes in Gemfile to change the rails version as 5.1.6


kelfink:
I wonder why it was still complaining about 5.2.  But fi this works, let's do it.

PR #4, opened 2018-05-01
Added redux-sagas

RamuKammagani-osi:


app/javascript/packs/application.js
```diff
@@ -7,4 +7,4 @@
 // To reference this file, add <%= javascript_pack_tag 'application' %> to the appropriate
 // layout file, like app/views/layouts/application.html.erb
 
-console.log('Hello World from Webpacker')
+console.log('Hello World from Webpacker');
```
kelfink: let's remove the console.log calls unless we need them.

app/javascript/_services/api/api.js
```diff
@@ -0,0 +1,8 @@
+import axios from 'axios';
+
+const api = axios.create({
+  baseURL: process.env.COG_APP_JS_API_URL || '/cognito/api/',
```
kelfink: Can we name this COUNTY_ADMIN_JS_API_URL so we're not referring to cognito ?

kelfink: and /county_admin/api as well

app/javascript/_services/api/api.test.js
```diff
@@ -0,0 +1,11 @@
+const ApiService = require('./api').default;
+
+describe('ApiService', () => {
+  it('has a timeout of 15000 seconds', () => {
+    expect(ApiService.defaults.timeout).toBe(15000);
+  });
+
+  it('has a baseUrl of /api/', () => {
+    expect(ApiService.defaults.baseURL).toBe('/cognito/api/');
```
kelfink: this doesn't line up with the it block mentioning /api/ .  And let's make it county-admin/api, or countyadmin/api

PR #10, opened 2018-05-09
Integrate with Mock Service for List of Users

RamuKammagani-osi:


app/javascript/home/home.js
```diff
@@ -0,0 +1,36 @@
...
+import { bindActionCreators } from 'redux';
+import { connect } from 'react-redux';
+import { selectUserRecords } from '../selectors/userListSelector';
+
+class Home extends React.Component {
+  componentDidMount() {
+    this.props.actions.fetchUsersActions();
+  }
+  render() {
+    return <CountyUsersListContainer userList={[this.props.userList]} />;
```
Sushmareddy-OSI: it can be {this.props.userList}

app/javascript/reducers/userListReducers.test.js
```diff
@@ -19,13 +19,16 @@ describe('reducer', () => {
   it('handles FETCH_USERS_API_CALL_SUCCESS', () => {
     const responseAction = {
       type: actionTypes.FETCH_USERS_API_CALL_SUCCESS,
-      userList: { 0: 'user1', 1: 'user2' },
+      userList: [{ 0: 'user1', 1: 'user2' }],
```
kelfink: this is an array of one element.  but it looks like you have two users.  The API response won't look like this, will it ?  [{ id: "key1", username: 'user1'}, { id: "key2", username: 'user2' }]. would be more representative.

app/javascript/reducers/userListReducers.test.js
```diff
@@ -19,13 +19,16 @@ describe('reducer', () => {
...
     };
     const state = { userList: null, fetching: true, error: null };
+
     expect(reducer(state, responseAction)).toEqual({
       fetchUserList: {
         fetching: false,
-        userList: { records: { 0: 'user1', 1: 'user2' }, XHRStatus: 'ready' },
+        userList: [
+          { records: [{ 0: 'user1', 1: 'user2' }], XHRStatus: 'ready' },
```
kelfink: as above, this seems like a strangely represented array.  A test value would be better as an actual array of multiple elements, and not the numerically indexed hash.

app/javascript/_services/users/user.service.test.js
```diff
@@ -3,8 +3,6 @@ import UserService from './user.service';
 jest.mock('../api');
 const ApiService = require('../api').default;
 
-const id = 'FeLX5t8aah';
-
 describe('UserService', () => {
   it('exists', () => {
     expect(!!UserService).toBeTruthy();
```
kelfink: The next test will clearly show that this exists.  It's good for TDD to have these temporarily, but otherwise I think it's just clutter.  Removing it also means we can remove the wrapping it 'exists' block.  less code is better.

app/javascript/reducers/userListReducers.test.js
```diff
@@ -19,13 +19,25 @@ describe('reducer', () => {
...
+        { id: 'key2', username: 'user2' },
+      ],
     };
     const state = { userList: null, fetching: true, error: null };
+
     expect(reducer(state, responseAction)).toEqual({
       fetchUserList: {
         fetching: false,
-        userList: { records: { 0: 'user1', 1: 'user2' }, XHRStatus: 'ready' },
+        userList: [
```
kelfink: an array of lists seems like a weird response shape.  Why would we ever get back more than one list of users?

app/javascript/reducers/userListReducers.test.js
```diff
@@ -19,13 +19,25 @@ describe('reducer', () => {
   it('handles FETCH_USERS_API_CALL_SUCCESS', () => {
     const responseAction = {
       type: actionTypes.FETCH_USERS_API_CALL_SUCCESS,
-      userList: { 0: 'user1', 1: 'user2' },
+      userList: [
+        { id: 'key1', username: 'user1' },
+        { id: 'key2', username: 'user2' },
+      ],
```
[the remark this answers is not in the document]
RamuKammagani-osi: 👍

app/javascript/home/home.js
```diff
@@ -0,0 +1,36 @@
...
+    return <CountyUsersListContainer userList={this.props.userList} />;
+  }
+}
+
+Home.propTypes = {
+  actions: PropTypes.object.isRequired,
+  userList: PropTypes.array,
+};
+
+function mapStateToProps(state, _ownProps) {
```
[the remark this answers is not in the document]
RamuKammagani-osi: will do, Thanks

app/javascript/views/usersList/UsersList.test.js
```diff
@@ -22,13 +22,8 @@ describe('UsersList', () => {
...
 
-  it('#displayList function defined', () => {
-    const instance = wrapper.instance();
-    expect(instance.displayList()).toBeDefined();
-  });
-
-  describe('#displayList output', () => {
-    it('displays table', () => {
+  describe('#UserList output', () => {
+    it('UserList', () => {
```
kelfink: it('contains table and headers')

kelfink:
retest this please

PR #14, opened 2018-05-14
fix full name order and sort functionality for last name

RamuKammagani-osi:


app/javascript/views/usersList/UsersList.jsx
```diff
@@ -22,10 +22,10 @@ const UserList = ({ userList }) => (
                 striped={true}
                 hover={true}
               >
-                <TableHeaderColumn dataField="first_name" dataSort width="100">
+                <TableHeaderColumn dataField="last_name" dataSort width="100">
                   Full Name
```
kelfink: why does the description Full Name not match the data field Last Name?  Can we sort by the computed full name, with last name and first name ?

PR #15, opened 2018-05-15
update symlink to web packs

kelfink:
Trying to fix pack references 404 in preint


kelfink:
Debugging Jenkins, docker, web pack symptoms in pre-int.

PR #17, opened 2018-05-16
Display User Details using mock service

RamuKammagani-osi:


app/javascript/sagas/getDetailsSaga.js
```diff
@@ -0,0 +1,29 @@
+import UserService from '../_services/users';
+import * as actionTypes from '../actions/actionTypes';
+import { takeLatest, call, put } from 'redux-saga/effects';
+
+const id = 'id';
+// worker saga: makes the api call when watcher saga sees the action
+export function* getDetails() {
+  try {
+    const details = yield call(UserService.fetchUserDetails, id);
```
kelfink: So this is a hard-coded 'fake' id for now?

app/javascript/selectors/detailSelector.js
```diff
@@ -0,0 +1,5 @@
+export const selectDetailRecords = state => {
+  console.log('usersDetails state', state);
```
kelfink: OK for now if we're still just expecting mock service, but ultimately this needs to be removed.

app/javascript/views/userDetail/UserDetail.jsx
```diff
@@ -5,74 +5,65 @@ import ShowField from '../../common/ShowField';
...
+const UserDetail = ({ details }) => (
   <div>
     <GlobalHeader profileName="County CWDS-Admin" profileId="profile.id" />
     <PageHeader pageTitle="Manage Users" button="" />
     <div className="container">
       <div className="row">
         <div className="col-md-12">
           <Cards
-            cardHeaderText="Manage Users: {county_name}"
+            cardHeaderText={'Manage Users: Sacramento'}
```
kelfink: Do we have access to an object that represents our current user (logged in) so we can find the county / region?  Or do we need to get this back from the API ?

Sushmareddy-OSI: +1

RamuKammagani-osi: Sure! I'll update that, Thanks

app/javascript/selectors/detailSelector.test.js
```diff
@@ -0,0 +1,35 @@
+import { selectDetailRecords } from './detailSelector';
+
+describe('selectors', () => {
+  describe('#selectUserRecords', () => {
+    it('selects the user records', () => {
```
Sushmareddy-OSI: change the description to select detail records

RamuKammagani-osi: Will do, thanks

PR #18, opened 2018-05-16
Selected user redirects to details page COG-63

Sushmareddy-OSI:
Added <a> tag to link UserDetail page from UserList, 
Concatenated LastName and FirstName and supplied to one column of a table as FullName

### DIFF
--- a/app/javascript/home/home.js
+++ b/app/javascript/home/home.js
@@ -13,7 +13,7 @@ class Home extends React.Component {
 
   link = (cell, row) => (
     <a href={`/user_details/${row.id}`}>
-      {row.last_name} {row.first_name}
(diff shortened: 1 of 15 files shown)


app/javascript/home/home.js
```diff
@@ -10,8 +10,19 @@ class Home extends React.Component {
   componentDidMount() {
     this.props.actions.fetchUsersActions();
   }
+
+  link = (cell, row) => (
```
RamuKammagani-osi: Better name for this function would have been great! Like formartName, fullNameFormat... etc.

Sushmareddy-OSI: agree!!

PR #23, opened 2018-05-21
Update Docker compose to get list of users.

RamuKammagani-osi:
Updating the docker-compose file to get list of users from actual implementation instead of calling the Mock Services from Perry


RamuKammagani-osi:
retest this please

RamuKammagani-osi:
retest this please

PR #24, opened 2018-05-22
Adding env variable prefix to rails assets path

ratneshraval:


ratneshraval:
@kelfink you can test on your computer by running 
`RAILS_RELATIVE_URL_ROOT=/cap bundle exec rails assets:precompile RAILS_ENV=production`

kelfink:
I think we can close this now, as the work was much bigger and ended up on another branch.

ratneshraval:
Agree

PR #25, opened 2018-05-23
Fix css links cog 10

kelfink:
A working solution to the CSS links not showing up in pre-int.  Manually adding the bootstrap, react-wood-duck, and react-bootstrap-tables CSS files to the docker image.


kelfink:
Successful changes get us into Integration.

PR #28, opened 2018-05-26
Adjust react links with rails relative root url

kelfink:
Adds rails relative url to links on list and detail pages

### DIFF
--- a/app/javascript/home/home.js
+++ b/app/javascript/home/home.js
@@ -12,7 +12,13 @@ class Home extends React.Component {
   }
 
   nameFormat = (cell, row) => (
-    <a href={`/user_details/${row.id}`}>
+    <a
+      href={`${[
+        process.env.RAILS_RELATIVE_URL_ROOT,
+        'user_details',
+        row.id,
+      ].join('/')}`}
+    >
       {row.last_name}, {row.first_name}
     </a>
   );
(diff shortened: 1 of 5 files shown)


app/javascript/views/usersList/UsersList.test.js
```diff
@@ -36,5 +36,9 @@ describe('UsersList', () => {
           .html()
       ).toContain('Dashboard');
     });
+
+    it('first link is pointed at dsashboardf', () => {
```
edgibbs: Looks like a typo with dshahboardf

app/javascript/home/home.js
```diff
@@ -12,7 +12,13 @@ class Home extends React.Component {
...
 
   nameFormat = (cell, row) => (
-    <a href={`/user_details/${row.id}`}>
+    <a
+      href={`${
+        process.env.RAILS_RELATIVE_URL_ROOT === '/'
+          ? '/'
+          : process.env.RAILS_RELATIVE_URL_ROOT + '/'
+      }${['user_details', row.id].join('/')}`}
+    >
```
edgibbs: Is this ternary actually tested for both sides, I think you'd have to mess with process.env to test.

kelfink: Havin trouble building a test for this class.  I'll have to check it out tomorrow.

PR #31, opened 2018-05-29
COG-50, DetaillsPage Edit mode functionality

Sushmareddy-OSI:


app/javascript/sagas/getDetailsSaga.test.js
```diff
@@ -1,11 +1,16 @@
...
+  getUserId,
+  currentPathname,
+} from './getDetailsSaga';
 import * as actionTypes from '../actions/actionTypes';
 import { takeLatest, call, put } from 'redux-saga/effects';
 
-const id = 'id';
-UserService.fetchUserDetails(id);
+// const id = 'id';
+// UserService.fetchUserDetails(id);
```
RamuKammagani-osi: can remove these lines.

app/javascript/common/Card.js
```diff
@@ -0,0 +1,100 @@
...
+  const classField = classNames(
+    cardbgcolor,
+    wrapContainer,
+    `col-lg-${columnLargeWidth}`,
+    `col-md-${columnMediumWidth}`,
+    `col-md-offset-${offsetMediumValue}`,
+    `col-xs-${columnXsmallWidth}`
+  );
+  let editClass = '';
+  props.cardActionButtons ? (editClass = 'edit') : (editClass = '');
```
kelfink: I find this syntax a bit deceptive.  editClass only gets changed once, but conditionally in one or the other place.   I'd prefer a single conditional assignment...

const editClass = props.cardActionButtons ?  'edit' : '';

app/javascript/common/Card.js
```diff
@@ -0,0 +1,100 @@
...
+          <span>{props.cardHeaderText}</span>
+          {props.cardHeaderButton && !props.cardActionButtons ? (
+            <Button
+              btnClassName="default pull-right"
+              btnName="Edit"
+              onClick={onEdit}
+            />
+          ) : (
+            ''
+          )}
```
kelfink: Let's change this technique:  https://reactjs.org/docs/conditional-rendering.html#inline-if-with-logical--operator

It is significantly simpler, because it gets rid of the else clause you've built up.  

Basically:
```
 + {props.cardHeaderButton && !props.cardActionButtons  && (
 +            <Button
 +   ...
 +                />
 +    )
 +  }
```
and you're done.   The && makes the component only get rendered

app/javascript/common/Card.js
```diff
@@ -0,0 +1,100 @@
...
+              <Button
+                btnClassName="primary"
+                disabled={enableSave}
+                btnName="save"
+                onClick={onSave}
+              />
+            </div>
+          ) : (
+            ''
+          )}
```
kelfink: same trick as above

app/javascript/common/Card.test.js
```diff
@@ -0,0 +1,76 @@
...
+    expect(
+      component
+        .find('span')
+        .at(0)
+        .props().children
+    ).toEqual(card.cardHeaderText);
+    expect(component.props().children).toEqual(card.children);
+  });
+
+  it('when cardHeaderButton is true edit button displays', () => {
```
kelfink: usually 'when' is an indicator to the developer that it's time to build a 'describe' block because there is a setup, like building a post, and inside that, you might have several it blocks. It's ok for now.

app/javascript/views/userDetail/UserDetail.jsx
```diff
@@ -1,101 +1,155 @@
...
-                    {' '}
-                    {details.start_date}
-                  </ShowField>
-                </div>
-                <div className="col-md-3">
-                  <ShowField label="End Date">{details.end_date}</ShowField>
-                </div>
+              {this.state.alert && (
+                <Alert alertClassName="success" faIcon="fa-check-circle">
+                  {'Your changes have been made successfuly'}
```
kelfink: Is this text in the designs?  I'd prefer:  'Your changes have been saved successfully'

app/javascript/views/userDetail/UserDetail.jsx
```diff
@@ -1,101 +1,155 @@
...
+                </Alert>
+              )}
+              Back to:{' '}
```
kelfink: doesn't this just spit out a space char?  Are you really asking for an &nbsp;   ?

app/javascript/views/userDetail/UserDetail.jsx
```diff
@@ -1,101 +1,155 @@
...
+                          <ShowField label="Email">{details.email}</ShowField>
+                        </div>
+                        <div className="col-md-3">
+                          <ShowField label="Office Phone Number">
+                            <span>{details.phone_number}</span>
+                          </ShowField>
+                        </div>
+                        <div className="col-md-3">
+                          <ShowField label="Start Date">
+                            {' '}
```
kelfink: same question about &nbsp;

app/javascript/common/Card.js
Sushmareddy-OSI: This component is from "react-wood-duck", I just did modifications that we need for our application. I would prefer to leave it for now. Further, this needs to be in react-wood-duck.

app/javascript/views/userDetail/UserDetail.jsx
Sushmareddy-OSI: I got this text from design!!

app/javascript/common/Card.js
Sushmareddy-OSI: Its again react-wood-duck

kelfink:
For sake of moving forward, I'm merging now.  Ramu requested a couple of comments removed, and that can be done in a further commit.

PR #33, opened 2018-05-30
Search Users by Last Name

RamuKammagani-osi:


app/javascript/home/home.js
```diff
@@ -8,8 +8,16 @@ import { connect } from 'react-redux';
 import { selectUserRecords } from '../selectors/userListSelector';
 
 class Home extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      searchKey: '',
+    };
+    this.handleTextChange = this.handleTextChange.bind(this);
```
Sushmareddy-OSI: No Need to bind.

app/controllers/api/user_list_controller.rb
```diff
@@ -2,8 +2,13 @@
 
 module Api
   class UserListController < ActionController::API
+    # def index
+    #   users = Users::UserRepository.new.get_users(session[:token])
+    #   render json: users
+    # end
+
```
Sushmareddy-OSI: can be removed

RamuKammagani-osi:
retest this please

RamuKammagani-osi:
retest this please

chadbratton:
retest this please

lib/infrastructure/http_service.rb
```diff
@@ -8,8 +8,8 @@ def initialize(base_url = Rails.configuration.micro_services['perry_api_base_url
       @base_url = base_url
     end
 
-    def get(url, token)
-      http_connection.get("#{url}?token=#{token}")
+    def get(url, last_name, token)
```
kelfink: We do NOT want to add these parameters into the http_service.  I believe this is being addressed by the next story.  Either this needs to be fixed now, or I need you to add a refactor story for the next sprint that undoes this.  The parameters here need to be generic, and the url with query string must be constructed by the caller.

lib/users/user_repository.rb
```diff
@@ -6,14 +6,14 @@ def initialize(http_service = Infrastructure::HttpService.new)
       @http_service = http_service
     end
 
-    def get_users(token)
-      response = @http_service.get('/perry/idm/users', token)
+    def get_users(last_name, token)
+      response = @http_service.get('/perry/idm/users', last_name, token)
```
kelfink: This is where the url needs to be constructed, by adding the last name onto the url string.  We might add logic to the http service which correctly adds the token whether the url has a query string already, or not.

lib/infrastructure/http_service.rb
RamuKammagani-osi: Right! I have already a story in next sprint, Will fix that as soon as possible, Thanks

lib/users/user_repository.rb
RamuKammagani-osi: 👍

kelfink:
With a couple of refactors to be done in the next sprint, this is OK to go.  We will undo changes to http_service and perform url manipulations in the calling code instead.

PR #35, opened 2018-05-31
Save to perry cog 45

kelfink:
This adds the browser-side calls to submit a PATCH to user details o the rails app.


app/javascript/_services/users/user.service.test.js
```diff
@@ -20,11 +20,39 @@ describe('UserService', () => {
...
+    });
 
     it('calls #fetchUsersDetails ApiService', () => {
       getSpy.mockReturnValue(Promise.resolve({}));
       UserService.fetchUserDetails(id);
       expect(getSpy).toHaveBeenCalledWith('/user_detail/id');
     });
   });
+
+  describe('#savehUserDetails', () => {
```
Sushmareddy-OSI: spellcheck "#saveUserDetails"

RamuKammagani-osi: +1

app/javascript/_services/users/user.service.test.js
```diff
@@ -20,11 +20,41 @@ describe('UserService', () => {
...
+    let patchSpy;
+
+    beforeEach(() => {
+      patchSpy = jest.spyOn(ApiService, 'patch');
+    });
+
+    it('calls #fetchUsersDetails ApiService', () => {
```
RamuKammagani-osi: #saveUserDetails

app/javascript/views/userDetail/UserDetail.jsx
```diff
@@ -37,6 +39,13 @@ export default class UserDetail extends Component {
     });
   };
 
+  onSaveDetails = () => {
+    const id = getUserId(currentPathname());
+    const { details } = this.state;
+    const response = UserService.saveUserDetails(id, details);
```
RamuKammagani-osi: Why UserService is called in UserDetail Component, Shouldn't be this in Saga?? I think onClick function should fire up an action and based on the action, patch request should happen! i guess that's gonna happen in Saga's. Isn't it??

kelfink: Due to the complexity I punted on saga last sprint in favor of just  getting it to go directly.  How do you feel about me adding a tech debt item to transform it to saga ?

RamuKammagani-osi: I'm okay with that!

RamuKammagani-osi:
retest this please

PR #39, opened 2018-06-06
Switch True/False to Active/Inactive in UI 

RamuKammagani-osi:


app/javascript/selectors/detailSelector.test.js
```diff
@@ -2,17 +2,44 @@ import { selectDetailRecords, permissionsList } from './detailSelector';
 
 describe('selectors', () => {
   describe('#selectDetailRecords', () => {
-    it('selects the user detail records', () => {
+    it('selects the user detail records when enabled is true', () => {
       const state = {
         fetchDetails: {
           details: {
-            records: { county_name: 'first', id: '12', name: 'third' },
+            records: {
```
kelfink: this should be an array shouldn't it?

app/javascript/selectors/detailSelector.test.js
```diff
@@ -2,17 +2,44 @@ import { selectDetailRecords, permissionsList } from './detailSelector';
...
+        enabled: 'Active',
+        name: 'third',
+      });
+    });
+
+    it('selects the user detail records when enabled is false', () => {
+      const state = {
+        fetchDetails: {
+          details: {
+            records: {
```
kelfink: array of records ?

kelfink: oh, never mind this is the one gotten details record.

app/javascript/selectors/detailSelector.test.js
kelfink: never mind.  this is the details fetch.  Just gets one record.

PR #51, opened 2018-06-12
Add missing specs to get full coverage

RamuKammagani-osi:


RamuKammagani-osi:
retest this please

kelfink:
retest this please

RamuKammagani-osi:
retest this please

kelfink:
retest this please

RamuKammagani-osi:
100% Coverage

screen shot 2018-06-13 at 11 25 58 am

RamuKammagani-osi:
retest this please

app/javascript/views/userDetail/UserDetail.jsx
```diff
@@ -46,6 +46,12 @@ export default class UserDetail extends Component {
     this.setState({ isEdit: false, alert: true, saveResponse: response });
   };
 
+  onClick = () => {
```
kelfink: How about renaming this onEditCancel ?  onClick is generic, and doesn't relate to the onCancel event it is called by

app/javascript/views/userDetail/UserDetail.jsx
```diff
@@ -140,9 +146,7 @@ UserDetail.propTypes = {
 };
 
 UserDetail.defaultProps = {
-  userListUrl: process.env.RAILS_RELATIVE_URL_ROOT
-    ? process.env.RAILS_RELATIVE_URL_ROOT
-    : '/',
+  userListUrl: process.env.RAILS_RELATIVE_URL_ROOT,
```
kelfink: This will make it critical that the environment variable be present.

app/javascript/views/userDetail/UserDetail.test.js
```diff
@@ -46,6 +46,14 @@ describe('UserDetail', () => {
         expect(instance.state.enableSave).toBe(false);
       });
     });
+
+    it('toggles the isEdit flag', () => {
+      const instance = wrapper.instance();
+      instance.onClick();
```
kelfink: more reason to not have onClick be the name of this method.  It makes me wonder what onClick is, and what's being clicked.  But we don't care about clicking.  We care that edit mode is canceled.

app/javascript/views/userDetail/UserDetail.jsx
RamuKammagani-osi: I still don't understand why do we need an default props! We already defined in props to receive value as string, So why do we need default props again??

RamuKammagani-osi: Here process.env.RAILS_RELATIVE_URL_ROOT is undefined  in local. So if it undefined then it should take as '/', when running in environments it receives the path, so i don't see any need to add a condition here??

kelfink: I believe it's good practice to have defaults set up.  If it doesn't appear we need it,  I'm good with removing it, however.

PR #55, opened 2018-06-21
COG-215 Add User Button to UserList page

Sushmareddy-OSI:
- [x] Refactor Card Component  (button names as props)

- [x] Change enableSave prop name to disableActionBtn

- [x] Add "+ ADD A USER"Button to UserList card


app/javascript/common/Card.js
```diff
@@ -3,16 +3,22 @@ import PropTypes from 'prop-types';
 import classNames from 'classnames';
 import { Button } from 'react-wood-duck';
 
-function renderCancelSaveButtons(props, onCancel, enableSave, onSave) {
+function renderCancelSaveButtons(
```
RamuKammagani-osi: can we rename this function name something like renderCardButtons(),  
renderCancelSaveButtons () is a bit confusing

kelfink:
Code Climate says we have to break up those longer methods

Sushmareddy-OSI:
I think, we already discussed about this. 
Those issues are from Card component which is a react-wood-duck component, down the road we will push that component to react-wood-duck. I don't think we need to refactor it now!!

kelfink:
We will ignore the code climate  warnings on Card, as we don't want to stray far from the react-wood-duck version which will ultimately be the new version

kelfink:
retest this please

PR #57, opened 2018-06-22
Add rails relative url to the route path.

kelfink:


kelfink:
still skipping the Card length issues

kelfink:
retest this please

PR #58, opened 2018-06-26
Cog 134 add ada acessibiliy testing using axe-matchers

kelfink:
Produces errors for violations in the acceptance test, which is currently pended.


app/javascript/common/Card.js
```diff
@@ -14,7 +14,11 @@ function renderCancelSaveButtons(
     props.cardActionButtons &&
     !props.cardHeaderButton && (
       <div className="pull-right">
-        <Button btnClassName="default" btnName="cancel" onClick={onCancel} />
+        <Button
+          btnClassName="default cancel"
```
RamuKammagani-osi: There is no class name with default cancel i guess  it should be default as of my understanding

spec/acceptance/user_list_page_spec.rb
```diff
@@ -2,19 +2,31 @@
...
 
     user = first_user_link
     puts "user link: #{user.text}"
     user.click
     page_is_user_details
+    # expect(page).to be_accessible
+
+    click_on('Edit')
+    # expect(page).to be_accessible
+    find('.cancel').click
```
kelfink: This 'cancel' references the class name 'cancel' I changed in 'Card.js'. Normally I'd use click_on('Cancel') but it didn't work like click_on('Edit') did.  We can try to rectify that later, when we aren't maintaining Card on our own.

PR #59, opened 2018-06-27
Cog 134 add ada acessibiliy testing

kelfink:


app/javascript/containers/detailsContainer.js
```diff
@@ -9,6 +9,16 @@ import { fetchPermissionsActions } from '../actions/permissionsActions';
 import { bindActionCreators } from 'redux';
 
 function mapStateToProps(state) {
+  console.log(
```
RamuKammagani-osi: Please remove comments

kelfink: Ah, no, this is for debugging.  It works locally, but not on pre-int.

RamuKammagani-osi: 👍

PR #64, opened 2018-06-28
Fix Search user list bug

RamuKammagani-osi:


spec/lib/users/user_repository_spec.rb
```diff
@@ -7,19 +7,32 @@ module Users
     let(:http_service) { instance_double('Infrastructure::HttpService') }
     let(:user_repository) { UserRepository.new(http_service) }
     let(:token) { 'sample_token' }
-    let(:last_name) { 'manzo' }
+    let(:params) { { 'lastName' => 'pop' } }
```
edgibbs: can this be new hash style 'lastName': 'pop'

spec/lib/users/user_repository_spec.rb
```diff
@@ -7,19 +7,32 @@ module Users
...
+            .with('/perry/idm/users', params, token)
             .and_return(response)
-          expect(user_repository.get_users(last_name, token)).to eq([])
+          expect(user_repository.get_users(params, token)).to eq([])
+        end
+      end
+
+      context 'with users' do
+        let(:parameter) { { 'lastName' => '' } }
+        it 'without params' do
```
edgibbs: 'returns users' maybe

spec/lib/users/user_repository_spec.rb
edgibbs: sounds like no so this is fine.

RamuKammagani-osi: 👍

PR #67, opened 2018-07-05
COG-296-CANCEL button in User Profile page does not appear to cancel the changes

Sushmareddy-OSI:


app/javascript/views/userDetail/UserDetail.jsx
```diff
@@ -52,9 +52,12 @@ export default class UserDetail extends Component {
   };
 
   onEditClick = () => {
-    this.state.isEdit === false
-      ? this.setState({ isEdit: true })
-      : this.setState({ isEdit: false });
+    this.setState({ isEdit: true });
```
kelfink: It was toggling, before (second click sets edit to false....). But I think the button gets disabled when we get into edit mode.  So this is good simplification, yes?

app/javascript/views/userDetail/UserDetail.jsx
```diff
@@ -52,9 +52,12 @@ export default class UserDetail extends Component {
...
+  };
+
+  onCancel = () => {
+    this.setState({ isEdit: false });
+    this.props.actions.fetchDetailsActions();
```
kelfink: Yes, go reload the record.  This actually answers my previous question. We need to have different actions on "cancel". which reloads, versus clicking edit, which does not.

app/javascript/views/userDetail/UserDetail.jsx
```diff
@@ -78,7 +81,7 @@ export default class UserDetail extends Component {
                 selectedPermissions={this.formattedPermissions(
                   this.state.details.permissions
                 )}
-                onCancel={this.onEditClick}
+                onCancel={this.onCancel}
```
kelfink: Right, and completing the shift, the cancel button now performs the reload and switch back to view mode.  good.

app/javascript/views/userDetail/UserDetail.test.js
```diff
@@ -46,8 +46,6 @@ describe('UserDetail', () => {
         const instance = wrapper.instance();
         instance.onEditClick();
         expect(instance.state.isEdit).toEqual(true);
```
kelfink: SO we don't expect it to toggle try/false anymore, but we lack a test of the new method onCancel

Sushmareddy-OSI: I tried testing onCancel but it throwing another error saying that action is not a function.

app/javascript/views/userDetail/UserDetail.jsx
kelfink: I don't see a unit test for onCancel.  Also we have an acceptance test which can now (after clicking cancel). show that the record gets restored.  That acceptance test can be a backlog item, but needs to be written up.  I'll do that.

kelfink:
We are skipping the unit test requested but will circle back to it with a technical debt.

PR #73, opened 2018-07-13
Cog 112 Display Office Name in User List

RamuKammagani-osi:
- [x] Added CSS style to fix the Row Width.

- [x] Change in Column name from End Date to Office Name


app/javascript/views/usersList/UsersList.test.js
```diff
@@ -116,6 +116,38 @@ describe('UsersList', () => {
...
+          .find('TableHeaderColumn')
+          .at(3)
+          .props().dataField
+      ).toBe('racfid');
+      expect(
+        wrapper
+          .find('TableHeaderColumn')
+          .at(4)
+          .props().dataField
+      ).toBe('office');
```
kelfink: good!  there should have been breaking tests without this.

RamuKammagani-osi: 👍

PR #76, opened 2018-07-20
Update react-router, use basename prop

eckdanny-osi:
Note: invoking like `RAILS_RELATIVE_ROOT_URL=whatever ./bin/webpack-dev-server` changes the output dir for packs. I dont think we want this, so using a basename prop for same effect. Tested with the plain 'ol rails "just in time" compilation (`rails s`) and in combination with webpack-dev-server


RamuKammagani-osi:
Jenkins build fails because  of lint errors!

eckdanny-osi:
This issue has been piece-wise solved by other PRs. closing.

PR #85, opened 2018-07-25
Show a RWD alert when non-200 paginated api res

eckdanny-osi:


eckdanny-osi:
screen shot 2018-07-25 at 11 47 58 am

eckdanny-osi:
unclear what was going on with commit history in this PR. I cleaned up and force pushed.

PR #93, opened 2018-07-30
Cog 375 accessibility page navigation by tabbing 

Sushmareddy-OSI:


app/javascript/views/usersList/Pagination.jsx
```diff
@@ -45,23 +45,19 @@ class Pagination extends PaginationRT {
...
-                    </option>
-                  ))}
-                </select>
-              </div>
-            }
+            <select
+              onBlur={_ => {}}
+              onChange={e => onPageSizeChange(Number(e.target.value))}
+              value={pageSize}
+              tabIndex={0}
```
eckdanny-osi: is a manual tabIndex actually needed here?

eckdanny-osi:
when did the wrapping `<div tabIndex={0}>` get introduced?

app/javascript/views/usersList/Pagination.jsx
Sushmareddy-OSI: Yes! needed, on tabbing its not highlighting the focus

Sushmareddy-OSI: on Friday, I tried fixing this issue by wrapping <div tabIndex={0}> , but later on when I tested in Safari it's not working as expected. Since changed to manual tabIndex

Sushmareddy-OSI: Yes! you are right , no need manual tabIndex

app/javascript/views/usersList/Pagination.jsx
```diff
@@ -45,23 +45,18 @@ class Pagination extends PaginationRT {
...
-              </div>
-            }
+            <select
+              onBlur={_ => {}}
+              onChange={e => onPageSizeChange(Number(e.target.value))}
+              value={pageSize}
+            >
+              {pageSizeOptions.map((option, i) => (
+                // eslint-disable-next-line react/no-array-index-key
+                <option key={i} value={option}>
```
eckdanny-osi: since our `options` are mutually exclusive, we _should_ be able to index by value and not iterator index... but this is non-blocking PR nit

PR #94, opened 2018-07-30
Use client-side routing

eckdanny-osi:


eckdanny-osi:
retest this please

eckdanny-osi:
```
Step 8/16 : COPY package.json yarn.lock /app/
 ---> Using cache
 ---> 830eb0a49d5a
Step 9/16 : RUN yarn install --production=false --non-interactive --frozen-lockfile
 ---> Using cache
 ---> cd5195c8dcec
```

Umm... This PR definitely introduces changes in [`package.json`](https://github.com/ca-cwds/county-admin/pull/94/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R61). Anyone have an idea why docker is using image layers from cache?

eckdanny-osi:
The following step: `bundle exec rails assets:precompile` runs yarn install all over again. With `NODE_ENV=production`, none of the `devDependencies` get installed :(. This is sad b/c they would have been present if not clobbered. 

The workaround [here](https://github.com/rails/webpacker/issues/1561) suggests moving critical `devDeps` to `deps`. This is a smelly solution, but let's see if it clears CI...

eckdanny-osi:
passed CI. The best thing probably would be to dig into the rake task `assets:precompile` and see if we can opt out of yarn clobbering dependencies. I mentioned in retro the necessarily long build steps in status checks... installing yarn deps multiple times is definitely contributing to that.

config/routes.rb
```diff
@@ -17,4 +17,9 @@
     get 'permissions_list', to: 'permissions_list#index'
     post 'add_user', to: 'add_user#index'
   end
+
+  # Single Page App home
+  get '*path', to: 'dashboard#index', constraints: ->(request) do
```
RamuKammagani-osi: Missing specs??

eckdanny-osi: +1 a spec would be nice. Where to assert behavior of the rails router? I'll write a spec in `dashboard` but that doesn't seem correct.

RamuKammagani-osi: As all the paths are routed with dashboard, adding specs in dashboard is only way we can think of at the moment!

RamuKammagani-osi: As all the paths are routed with dashboard, adding specs in dashboard is only way we can think of at the moment!

kelfink: Could we just test that it does to the root / path, which the unit tests will be configured to?  Or set the DASHBOARD_URL value we had previously and make sure the route matches that?  Either way I think this will be covered by acceptance tests (soon.)

PR #96, opened 2018-07-30
COG-195 Change Card header Label to County

RamuKammagani-osi:


RamuKammagani-osi:
Yes, Just a label change in all the cards.

PR #97, opened 2018-07-30
Update Version series to 83

Sushmareddy-OSI:


eckdanny-osi:
This should ideally be a ENV var in the CI host

PR #99, opened 2018-07-31
Default lists and searches to sort by last_name keyword

kelfink:


eckdanny-osi:
Odd failure in Acceptance Test.

PR #100, opened 2018-07-31
Use permissionsList KV map in show/edit views

eckdanny-osi:
screen shot 2018-07-31 at 3 14 52 pm
screen shot 2018-07-31 at 3 14 46 pm

### DIFF
--- a/app/javascript/selectors/detailSelector.js
+++ b/app/javascript/selectors/detailSelector.js
@@ -1,3 +1,5 @@
+import safeGet from 'lodash.get';
+
 export const selectDetailRecords = state => {
   const usersObject = state.fetchDetails ? state.fetchDetails.details : null;
   const userDetails = usersObject ? usersObject.records : {};
@@ -8,4 +10,4 @@ export const selectDetailRecords = state => {
 };
 
 export const permissionsList = state =>
-  state.fetchPermissions.permissions.permissions;
+  safeGet(state, 'fetchPermissions.permissions.permissions', []);
--- a/app/javascript/selectors/detailSelector.js
+++ b/app/javascript/selectors/detailSelector.js
@@ -9,9 +9,9 @@ export const selectDetailRecords = state => {
 
 export const permissionsList = state => {
   return state.fetchPermissions.permissions
-    ? state.fetchPermissions.permissions.permissions.map(list => ({
-        label: list,
-        value: list,
+    ? state.fetchPermissions.permissions.permissions.map(item => ({
+        label: item.replace(/-rollout$/, ''),
+        value: item,
       }))
     : [];
 };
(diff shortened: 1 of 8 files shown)


app/javascript/selectors/detailSelector.js
```diff
@@ -9,9 +9,9 @@ export const selectDetailRecords = state => {
 
 export const permissionsList = state => {
   return state.fetchPermissions.permissions
-    ? state.fetchPermissions.permissions.permissions.map(list => ({
-        label: list,
-        value: list,
+    ? state.fetchPermissions.permissions.permissions.map(item => ({
+        label: item.replace(/-rollout$/, ''),
```
RamuKammagani-osi: screen shot 2018-07-31 at 3 30 42 pm


@eckdanny-osi  I just cloned your branch and tried to run it. But i see the above error and it breaks the page when i hit add user button on the user list page! Am i missing something??

eckdanny-osi:
This PR should freeze until [COG-327](https://osi-cwds.atlassian.net/browse/COG-327) has shipped new images to docker hub. The java API (upstream dep) will provide `permissionsList: Array<{label: string, value: string}>` when complete. Waiting for those images to become available now

app/javascript/selectors/detailSelector.js
eckdanny-osi: should be
```diff
- ? state.fetchPermissions.permissions.permissions.map(item => ({
+ ? state.fetchPermissions.permissions.map(item => ({
```
but not addressing until COG-327 puts new images on docker hub

app/javascript/views/userDetail/UserDetailShow.jsx
```diff
@@ -54,7 +67,7 @@ const UserDetailShow = ({ details, onEdit }) => (
             </div>
             <div className="col-md-3">
               <ShowField label="Assigned Permissions">
-                {details.permissions}
+                {renderPermissions(details.permissions, permissionsList)}
```
Sushmareddy-OSI: @eckdanny-osi , I cloned and tested this branch. Noticed that when new user added with permission from {permessionsList} and saved. The assigned permissions are not viewed on show card.

eckdanny-osi: does a subsequent `GET /user/:userId` API request get made when returning to `show`?

Let's avoid managing a cache of mutated resources on the client. If PUT/POSTs don't yield the updated resource in the response body, then the resource should to be fetched with a GET all over again.

eckdanny-osi:
### 👀 Failing Jenkins CI status checks 👀 

If we depend on cognito for authentication, _all_ of our acceptance test will fail. We're in a tough spot rt now because we can't develop locally without a 3rd party network dependency since there is no mock cognito service for docker-compose. Isolation is broken => impacts tests _and_ dev-prod parity 🙁

eckdanny-osi:
Closing in favor of #103

PR #101, opened 2018-08-01
Cog 281 Terminology Change for RACFID to CWS Login 

Sushmareddy-OSI:


lib/users/verify_user.rb
```diff
@@ -37,6 +37,7 @@ class VerifyUser < Dry::Struct
     attribute :expires, Types::String.optional
     attribute :location, Types::String.optional
     attribute :pragma, Types::String.optional
+    attribute :error_code, Types::String.optional
```
RamuKammagani-osi: missing specs!

Sushmareddy-OSI: done!

eckdanny-osi:
@Sushmareddy-OSI #103 (merged) encapsulates the _tweaking_ we had to do to get Acceptance Tests passing again. Merge `master` into your PR please and we should get some helpful CI checks again.

app/javascript/_utils/errorCodes.js
```diff
@@ -0,0 +1,12 @@
+export function verificationError(errorCode, email, racfid) {
```
eckdanny-osi: I think the signature for this fn should be 

```ts
(errCode: String, options: Map<{ [key: String]: any }> = {}): String
```

positional arguments will get quite messy as more cases are written

Sushmareddy-OSI:
from Java API we are getting response as 
`{
  "verification_passed": false,
  "verification_message": "No user with RACFID: THOMALA is found in CWSCMS",
  "error_code": "CAP-003"
}`
we don't want the verification message using word `RACFID`. We are manipulating the  verification_message based on error_code.

app/javascript/_utils/errorCodes.js
kelfink: Looks like a small enough change.  Should help us future-proof the code.

app/javascript/views/addUser/AddVerifiedUser.test.js
```diff
@@ -84,14 +84,12 @@ describe('AddVerifiedUser', () => {
       const newUserDetails = {
         verification_passed: false,
         verification_message: 'No ID',
+        error_code: 'CAP',
       };
       wrapper = shallow(
         <AddVerifiedUser verifyNewUserDetails={newUserDetails} />
       );
       expect(wrapper.find('Alert').length).toBe(1);
-      expect(wrapper.find('Alert').props().children).toBe(
```
kelfink: is there no message anymore?  This had ben verifying the text of the message, right?

Sushmareddy-OSI: @kelfink, method verificationError() in _utils/error_codes where we are overriding the message by error_codes. we are not using the message that is generated by java API

kelfink:
Looks like this story is still necessary because the messages Perry spits out are not formatted for user consumption.  They are formatted for debugging or technical purposes. Users, however, are not expected to understand terms like RACFID.  SO the further formatting is needed.

PR #102, opened 2018-08-02
Add cognito login helper

eckdanny-osi:
provides a cognito-ready `login` helper method for acceptance tests. 

_NOTE:_ requires env vars be provided for

- `COGNITO_USERNAME`
- `COGNITO_PASSWORD`.

Applied to the tail of #100 and invoking with:

```bash
RAILS_RELATIVE_URL_ROOT=/cap \
COGNITO_USERNAME=*** \
COGNITO_PASSWORD=*** \
bundle exec rspec spec/acceptance
```

the acceptance tests all passed 👍

### DIFF
--- a/spec/acceptance/support/cognito_login_helper.rb
+++ b/spec/acceptance/support/cognito_login_helper.rb
@@ -1,8 +1,9 @@
 # frozen_string_literal: true
+
 module LoginHelper
-  def login()
+  def login
     visit ENV.fetch('RAILS_RELATIVE_URL_ROOT', '/')
-    return if not current_url.include?('amazoncognito.com')
+    return unless current_url.include?('amazoncognito.com')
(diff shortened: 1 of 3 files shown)


eckdanny-osi:
I imagine getting ENV_VARS distributed to all our different test runners may be a pain. I'd _much_ prefer doing that instead of setting default values to something out of the dev pool, but we can for the short term.

eckdanny-osi:
I suppose we should expect the CI to fail until we march the docker image tags forward. Perhaps we do that here?

eckdanny-osi:
Ugh the joys of playing guess-and-check with Jenkins and a 12minute CI cycle 😒

I don't currently have perms to view/edit env var bindings. (@kfries do you?) I submitted a request to DevOps for perms and opened #103 to encapsulate the pain.

eckdanny-osi:
Can't use cognito logins w/o localhost endpoint (for testing) or a valid https endpoint. We're going to have to do something different.

PR #104, opened 2018-08-03
Set from (offset) to zero when pageSize changes

eckdanny-osi:
There is a bug where we change page size, the `from` (offset) doesn't get wiped out.

ex:

 1. Say we have a 100 records in total
 1. Set `size` to something small (e.g.; 5)
 1. Scrub forward a few pages (e.g.; 4 of 20)
 1. Change the `size` to something big (e.g.; 100)
 1. 🐛

The query that went out is `{...params, size: 100, from: 40}`. 

The query that _should_ go out is `{...params, size: 100, from: 0}`


eckdanny-osi:
Getting kind of annoyed with codeclimate noise... Had to track down the issue type (_Avoid too many return statements ..._) <- umm... how many is too many?

I think this value is `4`, -which is awfully small for a reducer.

PR #105, opened 2018-08-03
Increase Test coverage on JS tests

RamuKammagani-osi:
Previously test coverage is about 86%


app/javascript/views/addUser/AddUserDetail.test.js
```diff
@@ -89,9 +116,11 @@ describe('AddUserDetail', () => {
           actions={{
             fetchDetailsActions: mockFetchDetailsActions,
           }}
+          id={'some_id'}
         />
       );
       const instance = wrapper.instance();
+      expect(mockFetchDetailsActions).not.toHaveBeenCalled();
```
kelfink: something should explain that *when xyz* then mockFetchDetailsActions should not get called.   Should be another describe, maybe nested here, or maybe a separate block.

PR #109, opened 2018-08-09
COG-422, Add user card Alignment fix

Sushmareddy-OSI:


kelfink:
Acceptance tests no longer seem hung up on this accessibility issue, so my block no longer applies.  There is other work to do on fixing accessibility.

PR #111, opened 2018-08-14
Sort by last_name then first_name case-insensitively

kelfink:
Add first_name on elastic search query to achieve last+first sorting


kelfink:
retest this please

Jenkinsfile
```diff
@@ -68,7 +68,7 @@ node(node_to_run_on()) {
         }
         stage('Acceptance Test') {
           sh "docker-compose up -d --build"
-          sh "sleep 120"
+          sh "sleep ${ACCEPTANCE_SLEEP_TIME}"
```
RamuKammagani-osi: Do we need to add more env_vars?? I think we have more than enough now! What you think??

RamuKammagani-osi: Sounds good! 👍

PR #112, opened 2018-08-15
COG-250 Display extension with phone number

RamuKammagani-osi:


app/javascript/_utils/formatDetails.js
```diff
@@ -0,0 +1,12 @@
+export function formatPhoneNumber(details) {
+  const phone = details.phone_number ? details.phone_number : '';
+  const ext =
+    details.phone_number &&
```
kelfink: we don't show extension if there's no phone number?  Would be good to have unit test coverage of that.

RamuKammagani-osi: Done!

app/javascript/_utils/formatDetails.js
```diff
@@ -1,12 +1,10 @@
-export function formatPhoneNumber(details) {
-  const phone = details.phone_number ? details.phone_number : '';
+/* eslint camelcase: 0 */
+
+export function formatPhoneNumber(phone_number, phone_extension_number) {
```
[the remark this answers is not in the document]
kelfink: Also you won't need the aslant camelcase exception called out anymore.

app/javascript/_utils/formatDetails.test.js
```diff
@@ -7,7 +7,9 @@ describe('formatPhoneNumber', () => {
         phone_number: '1234567890',
         phone_extension_number: '011',
       };
-      expect(formatPhoneNumber(details)).toEqual('1234567890 Ext 011');
```
kelfink: And if you use the restructuring I suggest above, then the calls all go back to the simpler call with (details). which is nice, because the callers aren't responsible for which bits to send, and the function is quickly eliminating the bits it doesn't need.

RamuKammagani-osi: Updated in new commit! Thanks

app/javascript/_utils/formatDetails.js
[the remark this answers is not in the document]
RamuKammagani-osi: Updated in new commit! Thanks

app/javascript/views/addUser/AddVerifiedUser.jsx
```diff
@@ -61,8 +61,8 @@ const AddNewUser = ({
                     <ShowField label="Office Phone Number">
                       <span>
                         {formatPhoneNumber(
-                          verifyNewUserDetails.user.phone_number,
-                          verifyNewUserDetails.user.phone_extension_number
+                          verifyNewUserDetails.user,
+                          verifyNewUserDetails.user
```
kelfink: huh?  Just pass it once.  I'm surprise this works

SashaKyz:
retest this please

PR #114, opened 2018-08-16
COG-301 Clicking on Edit, CMS Login Disappears

Nasir2712:
https://osi-cwds.atlassian.net/browse/COG-301


Nasir2712:
@kelfink test case added. please take a look

PR #122, opened 2018-08-21
COG-203, print profileName

Sushmareddy-OSI:


app/javascript/views/header/HeaderComponent.jsx
```diff
@@ -1,26 +1,24 @@
...
 
   render() {
     return (
       <div>
         <GlobalHeader
-          profileName="County CWDS-Admin"
-          profileId="profile.id"
-          logoutUrl="/logout"
+          profileName={this.props.fullName}
           logoutCallback={this.logoutCallback}
```
RamuKammagani-osi: Isn't it the staff id = profileId, if so, why to remove it?? just a question out of curiosity!

Sushmareddy-OSI: I don't see any use of this props. Thought of removing if we are not using...(cleaning code)

RamuKammagani-osi: 👍

PR #125, opened 2018-08-21
COG-194, save button disable on edit

Sushmareddy-OSI:


app/javascript/views/userDetail/UserDetail.test.js
```diff
@@ -74,6 +74,7 @@ describe('UserDetail', () => {
         const instance = wrapper.instance();
         instance.onEditClick();
         expect(instance.state.isEdit).toEqual(true);
+        expect(instance.state.disableActionBtn).toEqual(true);
```
RamuKammagani-osi: What about the alert? I expect one more expectation that the alert will be false when onEditClick function is called!

PR #131, opened 2018-08-22
COG-250 Display text Ext if no extension with phone

RamuKammagani-osi:


app/javascript/_utils/formatters.test.js
```diff
@@ -1,20 +1,39 @@
...
+        phone_number: '1112222333',
         phone_extension_number: '011',
       };
-      expect(formatPhoneExtension(details)).toEqual(' Ext 011');
+      expect(formatPhoneExtension(details)).toEqual('1112222333 Ext 011');
     });
   });
 
-  describe('When extension is not available ', () => {
+  describe('When phone  and extension does not exists ', () => {
```
Sushmareddy-OSI: It will be clear if it is something like " When Phone exists and No extension exists"

app/javascript/_utils/formatters.test.js
```diff
@@ -1,20 +1,39 @@
...
     it('returns just empty ', () => {
```
Sushmareddy-OSI: it is not returning empty

PR #134, opened 2018-08-22
Cog 419 trigger acceptance test run

kelfink:
Set up a separated acceptance test run.


spec/acceptance/user_edit_details_spec.rb
```diff
@@ -12,6 +12,7 @@
   scenario 'user_details edit/save' do
     login
     page_has_user_list_headers
+    sleep 2
```
edgibbs: Later refactoring target.

PR #135, opened 2018-08-23
COG-250 phone-format-fix

RamuKammagani-osi:


app/javascript/_utils/formatters.test.js
```diff
@@ -37,6 +41,28 @@ describe('formatPhoneExtension', () => {
...
+  describe('phoneNumberFormatter ', () => {
+    it('processes an empty string', () => {
+      expect(phoneNumberFormatter('')).toEqual(null);
+    });
+
+    it('should return formatted number when given a 10 digit number', () => {
+      expect(phoneNumberFormatter('1234567890')).toEqual('(123) 456-7890');
+    });
+
+    it('should return null when given a 5 digit number', () => {
```
Sushmareddy-OSI: I would prefer to be `should return null when given a less than 10 digit number`

app/javascript/_utils/formatters.test.js
```diff
@@ -37,6 +41,28 @@ describe('formatPhoneExtension', () => {
...
+      expect(phoneNumberFormatter('12345')).toEqual(null);
+    });
+
+    it('should return null when given a 11 digit number', () => {
```
Sushmareddy-OSI: I would prefer to be `should return null when given a greater than 10 digit number`

app/javascript/_utils/formatters.js
```diff
@@ -2,14 +2,29 @@
...
+        /(\d{3})(\d{3})(\d{4})/,
+        '($1) $2-$3'
+      );
+      return phoneNumber;
+    }
+  }
+  return null;
+};
+
 export function formatPhoneExtension({ phone_number, phone_extension_number }) {
```
Sushmareddy-OSI: Can we change the name of the method to `formatPhone`. Since, formatPhoneExtension seems like specific to extension but here it includes both phoneNumber and Extension.

PR #138, opened 2018-08-27
Add pending accessible ui tests

kelfink:
Adds pending tests on each page to test accessibility with spec matchers.  Output will show up in the accessibility test run.


kelfink:
retest this please

PR #140, opened 2018-08-28
Tech debt , cleaning up unwanted code

Sushmareddy-OSI:
- [x] Change the method name phoneNumberFormatter to formatPhoneNumber (consistant naming convention )
- [x] Move constants from userList and userDetails to constants folder
- [x] Change method name formatPhoneNumberExt to formatPhoneNumberWithExt
- [x] Move renderPermissions method as formatSelectedPermissions from userDetailshow to formatters folder
- [x] Remove formattedPermissions from userdetails.jsx and adduserdetails.jsx
- [x] Add saga's to save userDetails (edit and adduserEdit)
- [x] Rename headerComponent as GlobalHeader and move to common folder
- [x] Cleanup some specs
- [x] Replace switch case for verificationError Code mapping with Error message


app/javascript/views/addUser/AddUser.jsx
```diff
@@ -20,12 +20,22 @@ export default class AddUser extends Component {
...
       },
+      id: props.id,
+      details: props.details,
       errorMessage: {
         emailError: '',
       },
     };
   }
 
+  UNSAFE_componentWillReceiveProps(nextProps) {
```
kelfink: Seems like now is a good time to look into replacing this method completely.  Let's research what should replace componentWillReceiveProps

PR #142, opened 2018-08-30
Add specs to get better coverage

RamuKammagani-osi:
- [x] Added specs in search User saga 

- [x] Added specs for all actions

- [x] Add specs for Pagination

- [x] Increase test coverage


app/javascript/store/syncStore.test.js
```diff
@@ -12,8 +12,9 @@ describe('SyncStore', () => {
         removeItem: jest.fn(),
       };
     });
+    let key = 'MY_KEY';
```
kelfink: const ?

PR #143, opened 2018-08-30
COG-89 Administrators own account is Read Only

Nasir2712:
make the admin account ready only by disabling the edit button on user detail.


lib/perry/account.rb
```diff
@@ -19,5 +19,6 @@ class Account < Dry::Struct
     attribute :county_name, Types::String.optional
     attribute :roles, Types::Array.of(Types::String).optional
     attribute :privileges, Types::Array.of(Types::String).optional
+    attribute :userName, Types::String.optional
```
RamuKammagani-osi: Are you sure that it is userName?? i guess it might be user_name.

Nasir2712: ya, it is userName.

Nasir2712: yup this is what they are using.

app/javascript/selectors/detailSelector.js
```diff
@@ -11,3 +11,25 @@ export const selectDetailRecords = state => {
 
 export const permissionsList = state =>
   safeGet(state, 'fetchPermissions.permissions.permissions', []);
+
+const getAccountUserName = state =>
+  state.fetchAccount.account && state.fetchAccount.account.account.userName
+    ? state.fetchAccount.account.account.userName
+    : '';
+const getUserDetailRecordId = state => {
```
RamuKammagani-osi: shouldn't this be "getUserDetailRecords" instead of "getUserDetailRecordId" as this returns records!

Nasir2712: it returns userDetailRecordId, check line 26.

app/javascript/selectors/detailSelector.js
```diff
@@ -11,3 +11,25 @@ export const selectDetailRecords = state => {
...
+  const userDetailRecord =
+    state.fetchDetails.details !== null
+      ? state.fetchDetails.details.records
+      : '';
```
kelfink: records is an array or object, but '' is the default if details is null?  Should it be an {} or [] maybe ?

app/javascript/selectors/detailSelector.test.js
```diff
@@ -100,4 +104,137 @@ describe('selectors', () => {
...
+          fetchDetails: {
+            details: null,
+            fetching: false,
+            error: null,
+          },
+          fetchAccount: {
+            account: {
+              XHRStatus: 'ready',
+              account: {
+                userName: null,
```
kelfink: This state doesn't agree with the it block description.  only details null?  account username is null as well, which I think is the sam was the case below.

kelfink: Not sure we need both cases.

app/javascript/selectors/detailSelector.js
```diff
@@ -11,3 +11,25 @@ export const selectDetailRecords = state => {
...
+  const userDetailRecordId =
+    userDetailRecord !== '' ? state.fetchDetails.details.records.id : '';
+  return userDetailRecordId;
+};
+
+export const checkEditDisableBtn = state => {
+  const accountUserName = getAccountUserName(state);
+  const userDetailRecordId = getUserDetailRecordId(state);
+  if (!(accountUserName === '' || userDetailRecordId === '')) {
```
kelfink: I'd prefer ```if (accountUserName !== ''  && userDetailRecordId !== '') {```.  This eliminates the negating wrapper and reads more clearly.

kelfink: Also, the methods being called are using '' to represent null values.  Can they return '' when the originating value not null?  Maybe it doesn't matter

app/javascript/containers/detailsContainer.js
```diff
@@ -3,6 +3,7 @@ import UserDetail from '../views/userDetail/UserDetail';
 import {
   selectDetailRecords,
   permissionsList,
+  checkEditDisableBtn,
```
ratneshraval: I would like it it said `checkEditDisabledBtn`. More readable.

app/javascript/selectors/detailSelector.js
ratneshraval: Once you set else to null you can simply do ` accountUserName && userDetailRecordId `

app/javascript/selectors/detailSelector.test.js
Nasir2712: removed a case where username was null and details were null.

Nasir2712:
@kelfink I am not sure if the API is prohibiting this behavior in Perry. I think @sorych can answer your question.

PR #147, opened 2018-09-04
Add specs to get better coverage

RamuKammagani-osi:


app/javascript/views/addUser/AddUserDetail.test.js
```diff
@@ -95,11 +95,19 @@ describe('AddUserDetail', () => {
   });
 
   describe('#componentDidUpdate', () => {
-    it('passes along the props', () => {
+    it('When previous id and updated id is same, mockFetchDetailsActions is not called ', () => {
```
kelfink: How about:   it('When previous id and updated id is same, no fetching is performed'

PR #151, opened 2018-09-06
COG-421, Error message reflects "CWS Login" and "Email" when user did not click "VERIFY USER" button.

Sushmareddy-OSI:


app/javascript/sagas/validateNewUserSaga.js
```diff
@@ -14,7 +14,7 @@ export function* validateNewUser(action) {
     // dispatch a success action to the store with the new account details
     yield put({
       type: actionTypes.VALIDATE_NEW_USER_API_CALL_SUCCESS,
-      verifyUserDetails,
+      verifyUserDetails: { verifyUserDetails, verify: action.payload },
```
kelfink: I'd like to see this nested node renamed, since it's the name of the parent node as well.  It looks confusing to me, and will taker longer for anyone else to understand.   What if we use a name like 'responseDetails'. or 'verifiedUser'. to represent the node we get back from the API ?

kelfink:
This change is also a good time to add an acceptance test.  We can recreate the steps and validate that the obsolete error no longer shows.

app/javascript/sagas/validateNewUserSaga.js
RamuKammagani-osi: +1

PR #152, opened 2018-09-10
Update to secure base image

kelfink:
Updating to Nesh's secure version (minimal base). which will soon be a cads standard, to remove vulnerabilities.  also has yarn pre-installed.

Base is build by : https://github.com/ca-cwds/CALS/blob/development/docker/base/Dockerfile


docker/web/Dockerfile
```diff
@@ -1,11 +1,7 @@
-FROM ruby:2.5.1
+FROM ratneshraval/cals-base:0.7
 ENV ENVIRONMENT_REFRESH=2017-10-11 \
```
ratneshraval: Should you change the `ENVIRONMENT_REFRESH` ?

PR #155, opened 2018-09-12
COG-507 Display User status

RamuKammagani-osi:
- [x]  Add User Status field (Read Only) in User Detail/User Detail Edit Card

- [x]  Add formatters to display the description based on status value

- [x]  Rename the Labels in List/Show/Edit cards.

- [x]  Add tests for user status field and formatters


app/javascript/_utils/formatters.js
```diff
@@ -47,3 +47,33 @@ export function formatSelectedPermissions(assignedPermissions, permissionList) {
       .join(', ')
   );
 }
+
+export function userStatusFormatter(status) {
```
ratneshraval: This isn't just formatter right? More of user status code to text translation/conversion. Not critical to change it, although wanted to point out.

app/javascript/views/userDetail/UserDetailShow.jsx
```diff
@@ -65,16 +66,26 @@ const UserDetailShow = ({
           </div>
           <div className="row">
             <div className="col-md-5">
+              <ShowField label="User Status">
+                {details.status}
+                <div>
+                  <font color="#A9A9A9">
```
ratneshraval: Looks like this color code is being used in multiple places. Why not add it to stylesheets?

app/javascript/views/userDetail/UserDetailEdit.jsx
```diff
@@ -70,6 +74,25 @@ const UserDetailEdit = ({
...
+                  </font>
+                </div>
+              </ShowField>
+            </div>
+            <div className="col-md-3">
+              <DropDownField
+                id="dropdown1"
+                selectedOption={details.enabled}
+                options={STATUS}
+                label="Account Status"
```
ratneshraval: This and everywhere else, you don't need double quotes `"` for static strings.

app/javascript/views/userDetail/UserDetailShow.jsx
RamuKammagani-osi: Sure! Will do, Thanks

app/javascript/views/userDetail/UserDetailEdit.jsx
RamuKammagani-osi: 👍

app/javascript/common/GlobalHeader.test.js
```diff
@@ -7,6 +7,8 @@ describe('Header Component', () => {
   beforeEach(() => {
     const props = {
       fullName: 'lastname firstname',
+      logoutUrl:
+        'https://www.exampleGlobalHeader.com/HeaderComponent-12345_TESTING',
```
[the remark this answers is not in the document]
RamuKammagani-osi: yes!

app/javascript/_utils/codeToTextTranslator.test.js
```diff
@@ -0,0 +1,27 @@
...
+    );
+    expect(userStatusTranslator('UNKNOWN')).toEqual(
+      'User status is not known.'
+    );
+    expect(userStatusTranslator('RESET_REQUIRED')).toEqual(
+      'Need to reset user.'
+    );
+    expect(userStatusTranslator('FORCE_CHANGE_PASSWORD')).toEqual(
+      'User has never logged in.'
+    );
```
kelfink: How about a case where we receive a status we've not mapped, like 'NOVEL_STATUS'. coming in from the API ?  Should we return ''  ?  Or in fact maybe we should return the incoming value so the user can see that something is there?

app/javascript/_utils/codeToTextTranslator.js
```diff
@@ -0,0 +1,13 @@
+export function userStatusTranslator(status) {
+  const description = {
+    UNCONFIRMED: 'User has been created but not confirmed.',
+    CONFIRMED: 'User has been confirmed.',
+    ARCHIVED: 'User is no longer active.',
+    COMPROMISED: 'User is disabled due to a potential security threat.',
+    UNKNOWN: 'User status is not known.',
+    RESET_REQUIRED: 'Need to reset user.',
+    FORCE_CHANGE_PASSWORD: 'User has never logged in.',
+    undefined: '',
```
kelfink: this seems weird.  Does mapping undefined work ?  I don't think it's the same as not finding an unknown code value and returning that.  We probably need to know what to do if the incoming status cannot be mapped.   My suggestion before was to return description[status] || ''   but maybe it should be `return description[status] || status`

app/javascript/views/userDetail/UserDetailShow.jsx
```diff
@@ -65,16 +66,26 @@ const UserDetailShow = ({
           </div>
           <div className="row">
             <div className="col-md-5">
+              <ShowField label="User Status">
+                {details.status}
+                <div>
+                  <font className="value-text-color">
+                    {userStatusTranslator(details.status)}
+                  </font>
```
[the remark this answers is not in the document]
RamuKammagani-osi: Done! Thanks

app/javascript/_utils/codeToTextTranslator.js
```diff
@@ -0,0 +1,29 @@
+export function userStatusTranslator(status) {
+  switch (status) {
```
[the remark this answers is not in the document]
RamuKammagani-osi: Will do, Thanks

app/javascript/_utils/codeToTextTranslator.test.js
RamuKammagani-osi: Updated with other scenarios as well, Thank you for pointing it out!

app/javascript/_utils/codeToTextTranslator.js
RamuKammagani-osi: Done! updated in latest commit!

PR #156, opened 2018-09-12
Tech debt bug fix

Sushmareddy-OSI:


app/javascript/views/addUser/AddVerifiedUser.jsx
```diff
@@ -70,7 +70,7 @@ const AddNewUser = ({
             </Cards>
           </div>
         </div>
-      ) : verifiedUserDetails.verification_passed === false ? (
+      ) : verification_passed === false ? (
```
kelfink: should be able to refactor the to simply an else (:) clause without the comparison to ===

app/javascript/views/addUser/AddVerifiedUser.jsx
```diff
@@ -8,17 +8,17 @@ import { formatPhoneNumberWithExt } from '../../_utils/formatters';
...
+  let newUserDetails = verifyNewUserDetails
+    ? verifyNewUserDetails.verifiedUserDetails
+    : {};
+  let verification_passed = newUserDetails
+    ? newUserDetails.verification_passed
+    : '';
   return (
     <div>
-      {verifiedUserDetails.verification_passed ? (
+      {verification_passed === true ? (
```
kelfink: should just be able to ask if verification_passed ?  without comparison against ===

app/javascript/views/addUser/AddVerifiedUser.jsx
```diff
@@ -8,17 +8,17 @@ import { formatPhoneNumberWithExt } from '../../_utils/formatters';
...
-  email,
-  racfid,
-}) => {
-  const { verifiedUserDetails } = verifyNewUserDetails;
+const AddNewUser = ({ onCancel, onSave, verifyNewUserDetails }) => {
+  let verify = verifyNewUserDetails ? verifyNewUserDetails.verify : {};
+  let newUserDetails = verifyNewUserDetails
+    ? verifyNewUserDetails.verifiedUserDetails
+    : {};
+  let verification_passed = newUserDetails
```
kelfink: As long as you're assigning verification_passed with an expression used in a ternary, force it to be either true or false in this assignment.  No ''.  I think there' son use for an empty string here.

kelfink: I think these assignments (12-16) can be to 'const' variables

app/javascript/selectors/validateNewUserSelector.test.js
```diff
@@ -4,24 +4,16 @@ describe('selectors', () => {
...
         last_name: 'third',
+        roles: [],
+      };
+      expect(selectNewUserRecords(state)).toEqual({
         roles: ['CWS-worker'],
       });
     });
 
-    it('roles set to default value  when verifyNewUserDetails are not available', () => {
+    it('roles set to default value  when New User are not available', () => {
```
RamuKammagani-osi: Need to rephrase this sentence!

PR #157, opened 2018-09-13
COG-578-Update roles in county admin

sunilb0575:
**User Story:**

https://osi-cwds.atlassian.net/browse/COG-578

**Description:**

- User with one or all of the three (county-admin, state-admin, office-admin) roles is able to access CAP application.
- Old admin (with role CWS-admin) is not authorized to execute any CAP operations now (but he still can login as long as he exist in Cognito pool).


sunilb0575:
As of now, to get user lists both CWS-admin and COUNTY-admin should be available as roles. Waiting on @sorych to update dora. After his fix this failing test will pass. 
And yes I agree on having a flow when we have changes like this.

config/security.yml
```diff
@@ -1,2 +1,4 @@
 permitted_roles:
-      - 'CWS-admin'
+      - 'COUNTY-admin'
```
RamuKammagani-osi: Are you sure this is  County-admin or COUNTY-admin?? i remember @sorych mentioned something like County-admin. Just wanted to check

sunilb0575: @RamuKammagani-osi updated the roles as given in the user story.

PR #159, opened 2018-09-14
WIP Capture dora errors and report them usefully

kelfink:
Logging dora errors and returning a usable error message to the client should help in the future


Nasir2712:
i worked on the same issue as well, can you take a look at my PR, i actually created a middleware exception to catch every error response status.

kelfink:
> i worked on the same issue as well, can you take a look at my PR, i actually created a middleware exception to catch every error response status.

Looked at your PR.  Looks good.  I'll hold off on this PR until yours is merged, and see if there's any further need.

PR #160, opened 2018-09-14
COG-284 Tabbing exposes phantom button

RamuKammagani-osi:
- [x]  Refactor Card component

- [x] Add test cases 

- [x]  Update Verify/Add/User detail cards


RamuKammagani-osi:
retest this please

RamuKammagani-osi:
retest this please

RamuKammagani-osi:
screen shot 2018-09-18 at 1 57 00 pm
In this If two conditions satisfy it returns two buttons even if we need one, So  i just added a extra condition to return just one, we can pass props to card component whether we need one or two buttons.

RamuKammagani-osi:
@kelfink Sure, I'll see you at your desk then

kelfink:
should we remove this branch?

PR #161, opened 2018-09-17
Update chrome driver to 2.9 to try solving jenkins failure

kelfink:


Nasir2712:
In cals, the chrome driver version is set in the spec_helper file, maybe that could work. Can you check to see if you can add this line to the file?
```Chromedriver.set_version('2.38')```

PR #169, opened 2018-09-20
Code coverage improvement

RamuKammagani-osi:


app/javascript/views/userDetail/UserDetailEdit.test.js
```diff
@@ -96,6 +96,16 @@ describe('UserDetailEdit', () => {
       );
     });
 
+    it('MultiSelect', () => {
```
kelfink: better description, please

app/javascript/views/addUser/AddUser.test.js
```diff
@@ -89,7 +89,8 @@ describe('VerifyUser', () => {
 
   describe('#verifyUser()', () => {
     it('renders VerifyUser Component', () => {
-      const wrapper = shallow(<AddUser addUser={false} />);
```
kelfink: Was this here to create a specialized version of the wrapper?  Maybe we need to put this back but name is something else like wrapperAddUserFalse. for instance,  then use that I the rest of the it block.

PR #171, opened 2018-09-21
COG-499 Get Admin office ids and values 

RamuKammagani-osi:


app/javascript/selectors/accountSelectors.js
```diff
@@ -1,7 +1,5 @@
 export const selectCounty = state => {
```
kelfink: this method name no longer makes sense.   maybe selectAccount.   Or maybe selectLoggedInUser. ?

PR #176, opened 2018-09-24
COG-113 Filter users list by office

sunilb0575:
**User Story** 
https://osi-cwds.atlassian.net/browse/COG-113

**Description**
- User can select offices from the multi select dropdown.


sunilb0575:
I missed some test cases, working on it now. Please review rest of the code.

PR #178, opened 2018-09-27
COG-639 Integrate with API for User Details

RamuKammagani-osi:


RamuKammagani-osi:
Addressed in PR https://github.com/ca-cwds/county-admin/pull/179.

PR #182, opened 2018-09-28
COG-639 Disable edit button based on response from API

RamuKammagani-osi:
- [x] Refactor user detail services 

- [x]  Disable the edit button based on response from API

- [x] Add unit tests


app/javascript/containers/detailsContainer.js
```diff
@@ -17,12 +17,12 @@ import { bindActionCreators } from 'redux';
 
 function mapStateToProps(state) {
   return {
+    userEditOption: selectUserDetailObject(state),
```
Nasir2712: I would suggest you create a new selector for this, for example, something like getUserDetailEditableOption

app/javascript/selectors/detailSelector.js
```diff
@@ -1,36 +1,16 @@
 import safeGet from 'lodash.get';
```
Nasir2712: It would be much better if we handle all the necessary logic here in the selectors, just as we are doing for every other json from the API. If we create a new selector that will return the editable value, I think that should be very simpler. You could just return something like 
```!(state.fetchDetails.editable)``` -> this should return you the not of boolean or whatever the API is sending so that you can get rid of that mapper in the component.

File: app/javascript/views/userDetail/UserDetail.jsx [hunk 3]
```diff
@@ -134,7 +136,7 @@ export default class UserDetail extends Component {
                 details={this.state.details}
                 onEdit={this.onEditClick}
                 permissionsList={permissionsList}
-                disableEditBtn={this.props.disableEditBtn}
+                disableEditBtn={disableEditBtn(this.state.isUserEditable)}
```
Nasir2712: if you have everything in the selector, this could be just this.props.userEditOption

app/javascript/views/userDetail/UserDetail.jsx
```diff
@@ -15,6 +16,7 @@ export default class UserDetail extends Component {
       alert: false,
       disableActionBtn: true,
       details: props.details,
+      isUserEditable: props.userEditOption,
```
Nasir2712: you wouldn't need this.

app/javascript/selectors/detailSelector.test.js
```diff
@@ -1,10 +1,57 @@
...
+    });
+
+    it('selects the user detail object when fetchDetails exist', () => {
+      const state = {
+        fetchDetails: {},
+      };
+      expect(selectUserDetailObject(state)).toEqual(null);
+    });
+
+    it('selects the user detail object when details exist', () => {
```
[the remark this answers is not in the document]
RamuKammagani-osi: Done!

app/javascript/_utils/codeToTextTranslator.js
```diff
@@ -18,3 +18,19 @@ export function userStatusTranslator(status) {
...
+
+export function setAccountStatus(enabled) {
+  const status = {
+    true: 'Active',
+    false: 'Inactive',
+  };
+  return status[enabled];
+}
+
+export function disableEditBtn(editable) {
```
Nasir2712: you can also get rid of the method as well. if you handle everything in the selector.

app/javascript/_utils/codeToTextTranslator.js
```diff
@@ -18,3 +18,19 @@ export function userStatusTranslator(status) {
   };
   return userText[status] || '';
 }
+
+export function setAccountStatus(enabled) {
```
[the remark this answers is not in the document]
RamuKammagani-osi: Updated the function name

app/javascript/views/userDetail/UserDetailShow.test.js
```diff
@@ -102,7 +108,7 @@ describe('UserDetailEdit', () => {
       ).toEqual(expectedValue);
     });
 
-    it('checks disable edit button props', () => {
+    it('card disabled wprop will be set to true based on disableEditBtn', () => {
```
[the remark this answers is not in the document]
RamuKammagani-osi: Updated!

app/javascript/containers/detailsContainer.js

app/javascript/selectors/detailSelector.js
RamuKammagani-osi: Agreed, Updated!

app/javascript/views/userDetail/UserDetail.jsx
RamuKammagani-osi: Updated!

app/javascript/views/userDetail/UserDetail.jsx
RamuKammagani-osi: 👍

app/javascript/_utils/codeToTextTranslator.js
RamuKammagani-osi: Agreed, and Updated!

app/javascript/_utils/codeToTextTranslator.js
```diff
@@ -18,3 +18,19 @@ export function userStatusTranslator(status) {
...
+  };
+  return status[enabled];
+}
+
+export function disableEditBtn(editable) {
+  const disableBtn = {
+    true: false,
+    false: true,
+  };
+  return disableBtn[editable];
```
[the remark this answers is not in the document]
RamuKammagani-osi: I have updated the code, So i have removed the function and handled that logic in selector.

app/javascript/selectors/detailSelector.js
```diff
@@ -1,36 +1,19 @@
...
-
-export const checkEditDisabledBtn = state => {
-  const accountUserName = getAccountUserName(state);
-  const userDetailRecordId = getUserDetailRecordId(state);
-  if (accountUserName && userDetailRecordId) {
-    return accountUserName === userDetailRecordId;
-  }
-};
+export const checkEditDisable = state =>
+  !safeGet(state, 'fetchDetails.details.records.editable');
```
Nasir2712: will this handle null and undefined checks too?

RamuKammagani-osi: Yes, Added specs for those two scenarios.

Nasir2712: cool

app/javascript/views/userDetail/UserDetailShow.test.js
```diff
@@ -102,7 +108,7 @@ describe('UserDetailEdit', () => {
       ).toEqual(expectedValue);
     });
 
-    it('checks disable edit button props', () => {
+    it('sets card disabled prop to true based on disableEditBtn', () => {
```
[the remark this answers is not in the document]
RamuKammagani-osi: Updated!

PR #185, opened 2018-10-02
Fix autoloading user details

kelfink:
The UserDetails  class in lib/user.rb was occasionally being references before the User class had been found using auto-loading.  If you load the verify_user page before viewing a detail page, for instance, it tries to find the UserDetails class but can't because it's in the wrong file name.  Separating them fixes the issue.


app/javascript/views/usersList/UsersList.jsx
```diff
@@ -48,14 +48,17 @@ class UserList extends PureComponent {
...
   initialLoadQuery = loggedInUserAccount => {
-    return this.isOfficeAdmin(loggedInUserAccount)
-      ? [
-          {
-            field: 'office_ids',
-            value: loggedInUserAccount.admin_office_ids,
-          },
-        ]
-      : [];
+    if (this.isOfficeAdmin(loggedInUserAccount)) {
```
edgibbs: Ternary works, but I agree it's clearer here with the old fashioned if else

kelfink: I didn't want to call a mutating setOfficeList method in addition to the symmetric return I'm doing in the if and else, all in a ternary. so yeah.

PR #186, opened 2018-10-02
COG-637-Enable/disable search button

sunilb0575:
**User story:**
- https://osi-cwds.atlassian.net/browse/COG-637

**Description:**
- Endpoint for office list drop down is changed to 
```/perry/idm/admin-offices```
- Search button will remain disabled until the search criteria is changed.

### CODE REVIEW

File: app/javascript/views/usersList/UsersList.jsx [hunk 1]
```diff
@@ -130,13 +131,20 @@ class UserList extends PureComponent {
...
-    return lastNameSearch.value === nextSearch;
+    const { selectedOfficesList, nextSearch, query } = this.props;
+
+    const lastNameSearch = query.find(({ field }) => field === 'last_name');
+
+    const officeSearch = query.find(({ field }) => field === 'office_ids');
+
+    if (
+      _.isEqual(officeSearch.value.sort(), selectedOfficesList.sort()) &&
+      lastNameSearch.value === nextSearch
```
kelfink: reverse the order of these && clauses to shortcut the more expensive sorting and comparison/
Simply return the value of the &&'d clauses since the if and else blocks are just returning true / false respectively.

app/javascript/views/usersList/UsersList.jsx
```diff
@@ -18,6 +18,7 @@ import {
   lastLoginDate,
   officesListToOptions,
 } from '../../_constants/constants';
+import _ from 'lodash';
```
kelfink: I'd rather see us using the modularized https://www.npmjs.com/package/lodash.isequal. and not load all of lodash.

app/javascript/views/usersList/UsersList.jsx
```diff
@@ -237,6 +245,7 @@ class UserList extends PureComponent {
                         selectedOption={selectedOfficesList}
                         options={officesListToOptions(officesList)}
                         label="Filter by Office Name"
+                        placeholder={`(${officesList.length})`}
```
kelfink: isn't this line equivalent to :
`placeholder={officesList.length}`. ?

app/javascript/views/usersList/UsersList.test.js
```diff
@@ -212,13 +217,13 @@ describe('UsersList', () => {
...
           nextSearch="last_name_value"
+          selectedOfficesList={['somevalue']}
         />
       );
       expect(component.instance().isDisabledSearchBtn()).toEqual(true);
     });
 
-    it('returns false when search query is not passed', () => {
-      const query = [{}];
+    it('returns false when search query and entered search criteria are different', () => {
```
kelfink: Does this mean that we *always* pass a search query new ?  Because you have changed/removed this test case.

app/javascript/views/usersList/UsersList.jsx
kelfink: oh, those are literal on the screen.  I see.  Could those parens be rendered in the component instead of here?

kelfink: I see why now. Ok, accepted.

app/javascript/views/usersList/UsersList.test.js
kelfink: OK

sunilb0575:
jenkins, retest this please

kelfink:
retest this please

app/javascript/views/usersList/UsersList.jsx
```diff
@@ -130,13 +131,16 @@ class UserList extends PureComponent {
...
-    const lastNameSearch = query.find(
-      ({ field, value }) => field === 'last_name'
+    const { selectedOfficesList, nextSearch, query } = this.props;
+
+    const lastNameSearch = query.find(({ field }) => field === 'last_name');
+
+    const officeSearch = query.find(({ field }) => field === 'office_ids');
+
+    return (
+      lastNameSearch.value === nextSearch &&
```
kelfink: should be :` lastNameSearch && lastNameSearch.value === nextSearch &&`

PR #187, opened 2018-10-02
COG-645 Disable Resend Invite Button 

Sushmareddy-OSI:
- [x] Option for resend will be available in user detail view when status is 'Registration Incomplete'
- [x] Disable Resend Invite Button once it has been selected for that given session.
story : [`COG-645`](https://osi-cwds.atlassian.net/browse/COG-645)


app/javascript/selectors/resendEmailSelector.js
```diff
@@ -1,7 +1,27 @@
...
+  const userId = state.resendRegistrationEmail
+    ? state.resendRegistrationEmail.resendEmailUserId
+    : null;
+  return userId;
+};
+
+export const disableResendEmailButton = state => {
+  const userDetailsId = safeGet(state, 'fetchDetails.details.records.user.id');
+  const resendEmailUserId = selectResendEmailUserId(state);
+  let index = -1;
```
Nasir2712: Suggestion - check out lodash's ```_.every``` it will check for the condition and return a boolean, you won't need index logic.

PR #189, opened 2018-10-04
Cog 633 fix display of office

kelfink:
* fix office display by translating with the office list.
   Shows office name on add user, verify user, edit user, user list


kelfink:
I think the code climate number of functions can be ignored for the moment.  We'll prioritize cleanup I the next sprint.

app/javascript/views/userDetail/UserDetail.jsx
```diff
@@ -18,6 +18,8 @@ export default class UserDetail extends Component {
       resendEmailAlert: false,
       disableEditBtn: props.disableEditBtn,
       disableErrorAlert: false,
+      resendEmail: true,
+      officesList: props.officesList,
```
RamuKammagani-osi: We don't need extra state here as we are passing to edit/show cards as this.props.officeList.

Sushmareddy-OSI: I don't think we need this officeList in state, never used as state.

lib/users/user_details.rb
```diff
@@ -19,6 +19,7 @@ class UserDetails < Dry::Struct
     attribute :start_date, Types::String.optional
     attribute :end_date, Types::String.optional
     attribute :office, Types::String.optional
```
RamuKammagani-osi: You can remove this office attribute, as we not using this anymore.

spec/lib/users/user_details_spec.rb
```diff
@@ -16,6 +16,7 @@ module Users
       it { is_expected.to have_attribute(:start_date, Types::String.optional) }
       it { is_expected.to have_attribute(:end_date, Types::String.optional) }
       it { is_expected.to have_attribute(:office, Types::String.optional) }
```
RamuKammagani-osi: same here

app/javascript/_constants/constants.js
```diff
@@ -26,3 +26,14 @@ export const userStatusFormat = ({ enabled }) => {
 export const lastLoginDate = ({ last_login_date_time }) => {
   return checkDate(last_login_date_time);
 };
+
+export const translateOffice = (userObject, officesList) => {
```
Nasir2712: I honestly think this logic can be moved to the selector, and just the officeName props can be sent to the component. We can make use of the state and all the computational logic can be handled in the selectors.
This way it will be really easier for us to debug to just look at one place every time we debug.  
What do you think?

kelfink: Sounds good for a refactor.  Can we put it off until later so I can git thus merged ?

PR #191, opened 2018-10-08
Cog 461 values persists after logout

Sushmareddy-OSI:


kelfink:
This exposes and uses the deleteItem method, can calls it when the session is logged out.  Looks good.  Should have an acceptance test added.  Let's talk about that.

PR #192, opened 2018-10-08
Pass office_id for display on verify user component

kelfink:


lib/users/user_data.rb
```diff
@@ -0,0 +1,33 @@
...
+
+require 'dry-struct'
+require 'dry-types'
+
+module Types
+  include Dry::Types.module
+end
+
+module Users
+  class UserData < Dry::Struct
```
RamuKammagani-osi: Is it because of the initial loading issue we found  on verify user card?? Is that the reason for separating UserData from verify_user.rb ??

PR #193, opened 2018-10-09
Cog 684 get roles list

kelfink:
Added rails, react, redux functionality to load the dictionary data for Roles into browser props.    There is stub data produced which can be removed after the Perry roles service produces actual data instead of a 404 error.


lib/users/user_repository.rb
```diff
@@ -43,6 +43,15 @@ def get_offices_list(token)
       response.body { Offices.new }
     end
 
+    def get_roles_list(token)
+      response = @http_service.get('/perry/idm/roles', token)
+      if response.status == 404
```
RamuKammagani-osi: There will be new PR, that will replace this piece of code and just returns the response from API, As API service is not yet ready we are stubbing it to test some other parts on the FrontEnd.

lib/users/user_repository.rb
```diff
@@ -43,6 +43,15 @@ def get_offices_list(token)
       response.body { Offices.new }
     end
 
+    def get_roles_list(token)
```
money:
```
def get_roles_list(token)
       response = @http_service.get('/perry/idm/roles', token)
       return response.body { Roles.new } if response.status != 404

       [{ id: 'CWS-worker', name: 'CWS Worker (stub)' },
        { id: 'CALS-external-worker', name: 'CALS external worker (stub)' }]
end
```
How about this?

spec/acceptance/support/user_detail_page_helpers.rb
```diff
@@ -2,6 +2,8 @@
 
 module UserDetailPageHelper
   def page_is_user_details
+    # find forces a wait for the content to appear, while the expect does not.
```
money: Another option would be to wait for the AJAX to finish.  Not looking for this implementation, just a suggestion.  And you can look at the implementation of it here: https://github.com/ca-cwds/acceptance_testing/blob/master/spec/helpers/wait.rb

spec/controllers/api/roles_list_controller_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+require 'rails_helper'
+
+module Api
+  describe RolesListController do
+    describe '#index' do
+      let(:user_repository) { instance_double('User::UserRepository') }
+      let(:roles) { Users::Roles.new }
+
+      it 'has a route' do
+        expect(get: 'api/roles_list').to route_to(
```
money: There doesn't seem to be much value in this test.  You are better of testing if it returns a 200 status if the conditions are met and / or what the response body should include.

spec/controllers/api/roles_list_controller_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+          controller: 'api/roles_list',
+          action: 'index',
+          format: 'json'
+        )
+      end
+
+      it 'returns roles list' do
+        allow(Users::UserRepository).to receive(:new)
```
money: Maybe an oversight, but you can use the let variable from line here as well, like you have done on line 22

app/controllers/api/roles_list_controller.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+module Api
+  class RolesListController < ActionController::API
+    def index
+      token = session[:token]
```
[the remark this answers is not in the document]
kelfink: agreed.  I was able to one-line it

lib/users/roles.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+require 'dry-struct'
+require 'dry-types'
+
+module Types
```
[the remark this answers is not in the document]
kelfink: Done, but it's a bit weird to work off-pattern for dry-struct.  I've arranged a types file in the lib/users directory which we now refer to in all the model classes in lib/users. Shaves off a bit of code.

spec/acceptance/support/user_detail_page_helpers.rb
kelfink: I'll read up on it.  Just knowing which capybara methods cause waiting is enough, mostly.  The comment is just because it's not obvious why I'm doing both.  What I'd really like is to see a visible change (like Loading... going away). when relevant Ajax is completed.

lib/users/user_repository.rb
kelfink: I don't like the post-if clause with negatives like !=.   Actually by the time the parent story of this is complete, this whole 404 returning default values should go away.  Let's just leave this as-is.

spec/controllers/api/roles_list_controller_spec.rb
kelfink: You mean "allow(user_repository).to receive(:new). ?  Why would the object receive new?  This is is for creating an instance by calling a method on the class.

money: Perhaps consider using `allow_any_instance_of`(https://relishapp.com/rspec/rspec-mocks/docs/working-with-legacy-code/any-instance).  Either way, not a blocker.

Thanks for the clarification.

spec/controllers/api/roles_list_controller_spec.rb
kelfink: It's the test of the route.  Valid because we need to keep this stable.  I added a test for 200 status on getting the roles list.

kelfink:
retest this please

kelfink:
retest this please

PR #195, opened 2018-10-11
COG-649 Edit Roles 

RamuKammagani-osi:


kelfink:
retest this please

kelfink:
retest this please

kelfink:
retest this please

app/javascript/views/userDetail/UserDetail.jsx
```diff
@@ -47,15 +47,23 @@ export default class UserDetail extends Component {
...
-  onStatusChange = name => ({ value }) => {
+  handleOnRoleChange = name => ({ value }) => {
+    const { details } = this.state;
+    this.setState({
+      details: { ...details, [name]: [value] },
+      disableActionBtn: false,
+    });
+  };
+
+  handleOnAccountStatusChange = name => ({ value }) => {
```
NagaPradeepGajula-osi: these two methods here are identical doing the same thing. Can be refactored to avoid duplicate code.

sunilb0575:
jenkins, retest this please.

RamuKammagani-osi:
retest this please

app/javascript/_services/users/user.service.js
```diff
@@ -45,9 +45,22 @@ class UserService {
       });
     }
 
-    const data = { enabled: details['enabled'], permissions: permissions };
+    const roles = [];
+    if (Array.isArray(details.roles)) {
+      [].push.apply(roles, details['roles']);
```
Nasir2712: why are we using this logic? why cant we just do-
```const userData = {
roles: details['roles']
}```

File: app/javascript/views/userDetail/UserDetailEdit.jsx [hunk 3]
```diff
@@ -62,65 +90,78 @@ const UserDetailEdit = ({
...
             </div>
-            <div className="col-md-2">
-              <ShowField label="Last Login">
-                {checkDate(details.last_login_date_time)}
-              </ShowField>
+
+            <div className="col-md-4">
+              <DropDownField
+                id="dropdown1"
+                selectedOption={details.roles.toString()}
```
Nasir2712: why are we doing toString() here?

NagaPradeepGajula-osi:
@RamuKammagani-osi thanks for consolidating the methods.

app/javascript/views/userDetail/UserDetailEdit.jsx
RamuKammagani-osi: We are getting Response of roles as an Array, Based on the discussion with @sorych we should use 0 index value of the array, Most of the times it is an array with single value. So i took the array and converting it to string

Nasir2712: so you mean like this, details.roles[0]?

RamuKammagani-osi: yes

Nasir2712: so why are we not using just that?

RamuKammagani-osi: We can use that as well but need to write a separate event handler for that specific Dropdown, So instead of doing that, i used only one function so that we can optimize the code by converting it to the string.

app/javascript/_services/users/user.service.js
RamuKammagani-osi: Its an old code which we are not using that anymore, Thanks for pointing it out , i have updated the code in commit 3a18264

app/javascript/_services/users/user.service.test.js
```diff
@@ -114,15 +116,17 @@ describe('UserService', () => {
...
       patchSpy.mockReturnValue(Promise.resolve({}));
       const body = {
         enabled: true,
         permissions: [],
+        roles: [''],
       };
       UserService.saveUserDetails(id, {
         enabled: true,
-        permissions: '',
+        permissions: [],
```
kelfink: Good.  this goes toward expecting the state to always hold an array of permissions now, not a csv.

app/javascript/views/usersList/UsersList.jsx
```diff
@@ -270,7 +270,7 @@ class UserList extends PureComponent {
                         type="text"
                         value={this.props.nextSearch}
                         onChange={this.handleSearchChange}
-                        placeholder="search user by Last name"
+                        placeholder="Search user by Last name"
```
kelfink: Was this a design change?  Personally I think it should be "Search users by Last name" plural users

RamuKammagani-osi: No I just changed as its not starting with a capital letter. If it needs to be plural 'll update it.

app/javascript/views/userDetail/UserDetailEdit.jsx
RamuKammagani-osi: Nope, details.roles will always be an array as per the discussion with backend

PR #197, opened 2018-10-12
Add newrelic gem and configuration file

kelfink:
Basically the same commit used by CANS.   https://github.com/ca-cwds/cans/pull/207/files

I trimmed off unused portions of the newrelic.yml. since it only can be used in development, any environment the file is activated in will have newrelic disabled.   Deployers will replace the config file for any prod environment.


kelfink:
We can ignore the New Relic license issue mentioned by Code Climate

PR #198, opened 2018-10-15
Cog 295 Display Roles on User Detail View Page

Sushmareddy-OSI:


Sushmareddy-OSI:
retest this please

Sushmareddy-OSI:
retest this please

app/javascript/_utils/formatters.js
```diff
@@ -55,3 +55,15 @@ export function formatSelectedPermissions(assignedPermissions, permissionList) {
       .join(', ')
   );
 }
+
+export function formatSelectedRoles(assignedRoles, rolesList) {
```
Nasir2712: can you handle this logic in selectors? instead of wrapping this method inside the component?

kelfink:
retest this please

spec/lib/users/user_repository_spec.rb
```diff
@@ -117,8 +117,7 @@ module Users
             .with('/perry/idm/roles', token)
             .and_return(response)
           expect(user_repository.get_roles_list(token))
-            .to eq([{ id: 'CWS-worker', name: 'CWS Worker (stub)' },
-                    { id: 'CALS-external-worker', name: 'CALS external worker (stub)' }])
```
kelfink: yes.  we don't need the stub values any more.

app/javascript/_utils/formatters.test.js
```diff
@@ -120,3 +121,17 @@ describe('#formatSelectedPermissions', () => {
...
+  it('return first index of the array ', () => {
+    const list = [
+      { id: 'foo', name: 'FOO_DESC' },
+      { id: 'bar', name: 'BAR_DESC' },
+      { id: 'quo', name: 'QUO_DESC' },
+      { id: 'qux', name: 'QUX_DESC' },
+    ];
+    expect(formatSelectedRoles(['foo', 'bar'], list)).toEqual('FOO_DESC');
+    expect(formatSelectedRoles('', list)).toEqual('');
+    expect(formatSelectedRoles(['qux'], list)).toEqual('QUX_DESC');
```
kelfink: maybe a quick case for what happens when 'baz' is passed in but 'baz' is not in the list ?

Sushmareddy-OSI: a role id when it's not in the list will return undefined.

Sushmareddy-OSI: added unit test for the not listed option

PR #199, opened 2018-10-15
Cog 682 Full-Fill Possible role Options for Details Edit View

Sushmareddy-OSI:
Display Role Options based on user role


app/javascript/selectors/detailSelector.js
```diff
@@ -15,7 +18,7 @@ export const selectDetailRecords = state => {
 export const permissionsList = state =>
   safeGet(state, 'fetchPermissions.permissions.permissions', []);
 
-export const rolesList = state => safeGet(state, 'fetchRoles.roles.roles', []);
+export const rolesList = state => safeGet(state, 'fetchRoles.roles', []);
```
kelfink: Interesting.  I wonder why fetchPermissions has permissions nested in a deeper layer?  It must be in the redux or react Coe, as the network responses are the same shape
```
[{
    "name": "CANS-rollout",
    "description": "CANS"
}, {
    "name": "Facility-search-rollout",
    "description": "Facility Search \u0026 Profile"
},

```
pretty much the same as 
```
[{
    "id": "State-admin",
    "name": "State Administrator"
}, {
    "id": "County-admin",
    "name": "County Administrator"
},
```

app/javascript/selectors/detailSelector.test.js
```diff
@@ -166,13 +167,10 @@ describe('selectors', () => {
...
-            XHRStatus: 'ready',
-            roles: [
-              { id: 'foo-id', name: 'foo-name' },
-              { id: 'bar-id', name: 'bar-name' },
-            ],
-          },
+          roles: [
+            { id: 'foo-id', name: 'foo-name' },
+            { id: 'bar-id', name: 'bar-name' },
+          ],
```
kelfink: here's the shape change.  I like not having roles buried so deep.  Let's do this for permissions ?

lib/users/user_repository.rb
```diff
@@ -45,10 +45,7 @@ def get_offices_list(token)
 
     def get_roles_list(token)
       response = @http_service.get('/perry/idm/roles', token)
-      if response.status == 404
-        return [{ id: 'CWS-worker', name: 'CWS Worker (stub)' },
-                { id: 'CALS-external-worker', name: 'CALS external worker (stub)' }]
-      end
+      return [] if response.status == 404
```
kelfink: yeah we can stop the fake response now.

app/javascript/selectors/detailSelector.js
Sushmareddy-OSI: yes, in rolesReducers I did grab only roles from the state to avoid nested layer.

app/javascript/_utils/formatters.test.js
```diff
@@ -120,3 +121,17 @@ describe('#formatSelectedPermissions', () => {
     expect(formatSelectedPermissions(['qux'], list)).toEqual('QUX_DESC');
   });
 });
+
+describe('#formatSelectedRoles', () => {
+  it('return first index of the array ', () => {
+    const list = [
```
kelfink: Consider pulling this variable assignment out to a common variable list, then reference it in three small 'it' tests instead of three expect's.  If one expect fails here, the other two won't get tested.

app/javascript/_constants/constants.test.js
```diff
@@ -111,6 +112,23 @@ describe('helpers', () => {
...
   });
 
+  describe('#possibleRolesOptions', () => {
+    const rolesList = [
+      { id: 'role1', name: 'roleOne' },
+      { id: 'role2', name: 'roleTwo' },
+    ];
+    const possibleRoles = ['role1', 'role2'];
+    const expectedValue = [
+      { label: 'roleOne', value: 'roleOne' },
```
kelfink: Isn't there value (eventually sent in the update) supposed to be the id ?

app/javascript/_constants/constants.js
```diff
@@ -56,3 +56,21 @@ export const getOfficeTranslator = officesList => {
...
+  rolesList.forEach(role => {
+    possibleRoles.forEach(item => {
+      if (item === role.id) {
+        options.push(role.name);
+      }
+    });
+  });
+
+  return (
+    options &&
```
kelfink: I think you don't need to check options here.  It's always assigned above.

app/javascript/_constants/constants.js
```diff
@@ -56,3 +56,21 @@ export const getOfficeTranslator = officesList => {
   });
   return user => translateOfficeMap(user, officeMap);
 };
+
+export const possibleRolesOptions = (possibleRoles, rolesList) => {
```
Nasir2712: can you use this method in the selector?

app/javascript/views/userDetail/UserDetailEdit.jsx
```diff
@@ -62,29 +65,37 @@ const UserDetailEdit = ({
...
               <ShowField label="CWS Login">{details.racfid}</ShowField>
             </div>
-            <div className="col-md-2">
-              <ShowField label="Last Login">
-                {checkDate(details.last_login_date_time)}
-              </ShowField>
+            <div className="col-md-4">
+              <DropDownField
+                id="role"
+                options={possibleRolesOptions(possibleRoles, rolesList)}
```
Nasir2712: if you move this method to the selector, this will be just this.props.possibleRoles?

app/javascript/_constants/constants.js
Sushmareddy-OSI: if I move this to selector, I have to move following methods rolesList. That will be a refactor

PR #203, opened 2018-10-17
COG-648-Indication of user role name in list view

sunilb0575:
- https://osi-cwds.atlassian.net/browse/COG-648

- Fix for the above story, as role names should be shown in the list view.


sunilb0575:
jenkins, retest this place

sunilb0575:
jenkins, retest this please

sunilb0575:
jenkins, retest this please

app/javascript/_constants/constants.js
```diff
@@ -56,3 +56,15 @@ export const getOfficeTranslator = officesList => {
   });
   return user => translateOfficeMap(user, officeMap);
 };
+
+export function rolesTranslator(rolesID) {
+  const rolesName = {
```
RamuKammagani-osi: Instead of Constant, Can you look at the offices how are they mapped and displayed in list. So that if we see any other role which is not in the list, at least we can display the id.

kelfink: We have an office translation service.  why not use that ?

Sushmareddy-OSI: you can use formatSelectedRoles method in formaters.js file

PR #210, opened 2018-10-23
Display visible indicate the user could not be found

kelfink:
-
https://osi-cwds.atlassian.net/browse/COG-743
-  Failing to load a user record (valid problem or bug) should not show an empty page. Some indicator need to show the user that the user wasn't found.


app/javascript/views/userDetail/UserDetail.jsx
```diff
@@ -136,41 +139,52 @@ export default class UserDetail extends Component {
   };
 
   renderCards = permissionsList => {
+    console.log('state rendering cards', this.state);
```
Nasir2712: remove console.log

app/javascript/containers/detailsContainer.js
```diff
@@ -24,6 +25,7 @@ import { officesList } from '../selectors/userListSelector';
 
 function mapStateToProps(state) {
   return {
+    XHRStatus: fetchingStatus(state),
```
Nasir2712: we could just use fetching in the props as boolean.

app/javascript/selectors/detailSelector.js
```diff
@@ -22,3 +22,6 @@ export const rolesList = state => safeGet(state, 'fetchRoles.roles', []);
 
 export const checkEditDisable = state =>
   !safeGet(state, 'fetchDetails.details.records.editable');
+
+export const fetchingStatus = state =>
```
Nasir2712: we probably won't need this.

app/javascript/views/userDetail/UserDetail.jsx
```diff
@@ -201,7 +215,13 @@ export default class UserDetail extends Component {
             {this.state.disableErrorAlert &&
               userDetailError && <ErrorMessage error={userDetailError} />}
           </div>
-          {this.renderCards(permissionsList)}
+          {this.state.XHRStatus === 'ready' ? (
```
Nasir2712: this can be wrapped inside a method to avoid conditional rendering.

kelfink: Yeah I can wrap it inside renderCards itself

app/javascript/selectors/detailSelector.js
kelfink: Fetching is whether it's in process.  XHRStatus is whether it was successful

Nasir2712: Ya, wouldn't it make sense to show the user the page is loading and then display the results if they exist or not?

kelfink: ^ that's what this fix does..... "Loading..." shows, then the record.  Or "User could not be found" if we're not in the process of fetching and there's no detail.

Nasir2712: fetching means it's in the process, so shouldn't that indicate loading card?

kelfink: While the saga is running, it shows "Loading..."

app/javascript/containers/detailsContainer.js
kelfink: the fetching boolean just tells us if it is actively fetching, not whether it was successful when done

app/javascript/selectors/detailSelector.js
Nasir2712: had an offline conversation, explanation understood why another state is needd. Thanks for the clarification.

app/javascript/views/userDetail/UserDetail.jsx
```diff
@@ -138,41 +141,56 @@ export default class UserDetail extends Component {
   };
 
   renderCards = permissionsList => {
-    return (
+    console.log(this.state);
```
Nasir2712: this needs to be removed

app/javascript/views/userDetail/UserDetail.jsx
```diff
@@ -138,41 +141,56 @@ export default class UserDetail extends Component {
...
-                disableEditBtn={this.props.disableEditBtn}
-                officesList={this.props.officesList}
-                rolesList={this.props.rolesList}
-              />
-            )}
-          </div>
-        ) : (
-          ''
-        )}
+        <div>
```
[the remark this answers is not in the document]
kelfink: Good question.  Don't know.  Removed.  Looks like it was invisible anyway.

app/javascript/views/userDetail/UserDetail.jsx
sunilb0575: Also please remove this console.log.

PR #211, opened 2018-10-24
DoNotMerge - COG-764-Enable/disable roles dropdown

sunilb0575:
- https://osi-cwds.atlassian.net/browse/COG-764

- Docker image with the contract change `4.0.1_1069-RC`
- Below is the contract change
`
{"edit_details": {
      "editable": true,
      "roles" : {
        "editable" : true,
        "possible_values": [ "string"]
      }
    }
  "user": { }
}
`


app/javascript/selectors/detailSelector.test.js
```diff
@@ -263,4 +277,82 @@ describe('selectors', () => {
       expect(checkEditDisable(state)).toEqual(true);
     });
   });
+
+  describe('#disableRolesDropDown', () => {
+    it('return true when editable is false', () => {
+      const state = {
```
kelfink: To minimize lines added to this large file, let's pull this const declaration up into a function in the describe block, which takes as one parameter the value of 'editable'.  I'm doing that in may PR, and it saved dozens of lines and looks cleaner as well, I think.
```suggestion
      const getFetchDetailEditableState = editable => {
      return {
        fetchDetails: {
          details: {
            XHRStatus: 'ready',
            records: {
              editable: editable,
              user: {},
            },
          },
        },
      };
    };
    ...
    const state = getFetchDetailEditableState(false);  // or null or true or undefined
```

app/javascript/selectors/detailSelector.test.js
```diff
@@ -263,4 +277,82 @@ describe('selectors', () => {
...
+              user: {},
+            },
+          },
+        },
+      };
+      expect(disableRolesDropDown(state)).toEqual(true);
+    });
+
+    it('return the false if editable is true', () => {
+      const state = {
```
```suggestion
      const state = getFetchDetailEditableState(true);
```

RamuKammagani-osi:
@sunilb0575 Can you please remove the Tag DONOTMERGE from the title

PR #213, opened 2018-10-25
COG-744-UsersList Updates when added new users / updated existing users

Sushmareddy-OSI:
-COG-744-Manage Users list does not reflect changes immediately (> 18 minutes sometimes)

- UserList Updates with new added users and updated details.


app/javascript/views/usersList/UsersList.test.js
```diff
@@ -234,34 +239,61 @@ describe('UsersList', () => {
...
+          lastName="some_value"
+          officeNames={['north']}
+          inputData={{ officeNames: ['north'] }}
+          loggedInUserAccount={{ county_name: 'SomeCountyName' }}
           selectedOfficesList={['somevalue']}
         />
       );
     });
 
+    it('', () => {
```
Nasir2712: can you update this?

app/javascript/views/usersList/UsersList.test.js
```diff
@@ -234,34 +239,61 @@ describe('UsersList', () => {
...
+      comp.instance().componentDidMount();
+      expect(mockSetSearch).toHaveBeenCalledWith([
+        { field: 'last_name', value: 'some_value' },
+        { field: 'office_ids', value: ['north'] },
+      ]);
+    });
```
Nasir2712: componentDIdMount also has one more action about setPage, do we need that in expect statement as well?

Sushmareddy-OSI: Yes we need.

app/javascript/views/usersList/UsersList.test.js
```diff
@@ -318,7 +351,8 @@ describe('UsersList', () => {
           lastName="some_value"
           officeNames={['north']}
           inputData={{ officeNames: ['north'] }}
-        />
+        />,
+        { disableLifecycleMethods: true }
```
RamuKammagani-osi: i think you don't need to disable the lifecycle method here

PR #215, opened 2018-10-29
Updated add-user test to clean out old conflicting users

kelfink:
- Came up while cleaning up acceptance tests.  Failures can leave the one user we add still active, which breaks the next test run, due to only allowing one active of each RACFID

-

* Scroll through the user list deactivating any of the 'Auto' users and additionally testing that 'next' works to scroll after search.
* Also fixed a reported devops problem where our license-listing task fails because Gemfile was not listing the required gem.


RamuKammagani-osi:
@kelfink  I tried to run the acceptance tests again integration environment with this branch. I see 3 failures and sometimes 2 failures, am i missing something?? I would like to see if everything passes??

kelfink:
> @kelfink I tried to run the acceptance tests against integration environment with this branch. I see 3 failures and sometimes 2 failures, am i missing something?? I would like to see if everything passes??

I found capybara setting of empty string in the search last name field was not triggering the onChange, thus the search button sometimes remains disabled.  Also cleaned up some inconsistent errors with the clicking of user rows that might not be present on the list page. It all ran successfully in my local environment after that.

PR #216, opened 2018-10-29
COG-713 Fix permissions displaying 0 instead of empty.

RamuKammagani-osi:
- https://osi-cwds.atlassian.net/browse/COG-713

- Fixes the Permissions displaying '0' instead of empty when no permissions exists.


app/javascript/_utils/formatters.js
```diff
@@ -47,7 +47,7 @@ export function formatSelectedPermissions(assignedPermissions, permissionList) {
   if (!Array.isArray(assignedPermissions)) return '';
   return (
     assignedPermissions &&
-    assignedPermissions.length &&
+    assignedPermissions.length !== 0 &&
```
kelfink: Curious what went wrong with the .length.  Also, I personally prefer to see "x > 0" rather than x !== 0 when x cannot be negative ever.  It's safe and ready more smoothly, imo.

RamuKammagani-osi: if it has the zero index then its returning the 0, thats why i added one more check to return until it is greater than 0.

kelfink: Hmm.  example ?

RamuKammagani-osi: When the assignedPermissions has [] then its returning 0.

RamuKammagani-osi: Updated as x > 0,

app/javascript/_utils/formatters.js
```diff
@@ -47,7 +47,7 @@ export function formatSelectedPermissions(assignedPermissions, permissionList) {
   if (!Array.isArray(assignedPermissions)) return '';
   return (
     assignedPermissions &&
-    assignedPermissions.length &&
+    assignedPermissions.length > 0 &&
```
kelfink: Ah.  you're avoiding weird marshaling problems here.  the aP.length expression returns 0 and that's not truthy, so that gets returned, without marshaling, as 0.  This is the bug itself.  0 displayed to the end user.  Good fix.

RamuKammagani-osi: 👍

PR #217, opened 2018-10-29
COG-854 Refactor User Detail Page

Nasir2712:
- https://osi-cwds.atlassian.net/browse/COG-854

- refactored user detail page, to move everything except resendEmailAlert out from the state to reducer.


app/javascript/views/userDetail/UserDetailEdit.jsx
```diff
@@ -129,7 +129,7 @@ const UserDetailEdit = ({
                   selectedOption={details.enabled}
                   options={STATUS}
                   label="Account Status"
-                  onChange={onStatusChange}
+                  onChange={event => onStatusChange('enabled', event.value)}
```
RamuKammagani-osi: Missing specs for the onChange!

app/javascript/views/userDetail/UserDetailEdit.jsx
```diff
@@ -74,7 +74,7 @@ const UserDetailEdit = ({
                 selectedOption={details.roles.toString()}
                 id="RolesDropDown"
                 label="Role"
-                onChange={onRoleChange}
+                onChange={event => onRoleChange('roles', [event.value])}
```
RamuKammagani-osi: Missing specs for the onChange!

app/javascript/views/userDetail/UserDetail.test.js
```diff
@@ -216,11 +215,11 @@ describe('UserDetail', () => {
...
 
       it('renders card with text indicating no user found', () => {
-        wrapper.setState({
-          isEdit: true,
-          XHRStatus: 'ready',
+        wrapper.setProps({
+          details: {},
         });
         expect(wrapper.find('Cards').length).toBe(1);
+        console.log(wrapper.find('Cards').debug());
```
RamuKammagani-osi: remove console.log

app/javascript/views/userDetail/UserDetail.jsx
```diff
@@ -151,41 +115,50 @@ export default class UserDetail extends Component {
...
+    permissionsList,
+    XHRStatus,
+    details,
+    disableResendEmailButton,
+    officesList,
+    rolesList,
+    possibleRoles,
+    isRolesDisabled
+  ) => {
+    return XHRStatus !== 'ready' ? (
```
RamuKammagani-osi: Can you add specs when XHRStatus is not ready, and when it is ready?

app/javascript/reducers/detailsReducers.js
```diff
@@ -52,6 +63,7 @@ function fetchDetails(
         fetching: false,
         saveDetailsError: null,
         isEdit: false,
+        alert: true,
```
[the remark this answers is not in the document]
kelfink: agreed.  too much use around here, in general, of the word alert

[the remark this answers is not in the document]
Nasir2712: well, that was the point of the refactor, that alert appears only when you save the form and that's all it does. I am trying to keep alert as it is because, if in future if we need to have a separate alert for something else, then we can rename this to something more specific.

app/javascript/views/userDetail/UserDetailEdit.jsx
```diff
@@ -139,10 +139,13 @@ const UserDetailEdit = ({
                   options={permissionListToOptions(permissionsList)}
                   label="Assigned Permissions"
                   onChange={selectedOptions =>
-                    onPermissionChange(selectedOptions.split(','))
+                    onPermissionChange(
+                      'permissions',
```
kelfink: I don't like adding a string parameter to a method onPermissionChange that tells it it's dealing with 'permissions'.  Shouldn't it be obvious ?

app/javascript/reducers/detailsReducers.js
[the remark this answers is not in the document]
kelfink: Maybe a verb as part of this, like 'showAlert'. displayAlert ??

app/javascript/views/userDetail/UserDetailEdit.jsx
kelfink: Ok, it's because the function actually passed down is generic.  But it seems like we should have that function return another function given the name 'permission', since it's really a configuration of what we passed.  We should have examples of that pattern around here some place...

app/javascript/views/userDetail/UserDetail.jsx
```diff
@@ -151,41 +103,52 @@ export default class UserDetail extends Component {
...
               <UserDetailEdit
-                details={this.state.details}
-                selectedPermissions={this.state.details.permissions}
+                details={details}
+                selectedPermissions={details.permissions}
                 onCancel={this.onCancel}
                 onSave={this.onSaveDetails}
-                onStatusChange={this.handleDropDownChange('enabled')}
-                onPermissionChange={this.handleOnPermissionChange}
-                onRoleChange={this.handleDropDownChange('roles')}
```
kelfink: here's the pattern I was thinking of, where you configure the function in the creation of the component, so its use is simpler late on....  why change this?

app/javascript/common/DropDown.js
```diff
@@ -41,7 +41,7 @@ class DropDown extends React.Component {
 DropDown.propTypes = {
   options: PropTypes.array,
   id: PropTypes.string,
-  selectedOption: PropTypes.array,
+  selectedOption: PropTypes.oneOfType([PropTypes.string, PropTypes.array]),
```
sunilb0575: There is propType error here when selectedOption is boolean.

app/javascript/reducers/detailsReducers.js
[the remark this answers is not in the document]
Nasir2712: sure can change this displayAlert.

PR #219, opened 2018-10-31
Cog 794 tech debt move methods to selectors

Sushmareddy-OSI:
[COG-794](https://osi-cwds.atlassian.net/browse/COG-794) 

- Refactor the code 
- Methods Moved to selectors (detailsSelector.js)
 (formatDate, accountStatus, permissionList, formatSelectedPermissions, rolesList, possibleRolesFormatter, codeToTextFormatters)


app/javascript/selectors/detailSelector.js
```diff
@@ -32,3 +56,41 @@ export const disableRolesDropDown = state =>
 
 export const fetchingStatus = state =>
   safeGet(state, 'fetchDetails.details.XHRStatus');
+
+export const selectStartDate = state => {
```
RamuKammagani-osi: Can we keep this  method in formatters itself, it's also used in verified user card, also keeping the method generic will be better so that we can use this method for any date fields in future. I don't think we need a selector for start date! thoughts??

Sushmareddy-OSI: good thought to maintain date formatter as a generic method, but I would prefer to add generic method also in selector rather than in formatters. 
team suggestions ??

kelfink: Maybe the selector should call  a generic formatter method so that selector is specifically but the logic is generically available for others

app/javascript/selectors/detailSelector.test.js
```diff
@@ -58,6 +62,56 @@ describe('selectors', () => {
...
+      };
+    };
+    describe('When date exists ', () => {
+      it('with different format', () => {
+        const state = getStartDate('2001-09-01');
+        expect(selectStartDate(state)).toEqual('09/01/2001');
+      });
+    });
+
+    describe('When date not exists ', () => {
```
```suggestion
    describe('when date is an empty string ', () => {
```

app/javascript/selectors/detailSelector.test.js
```diff
@@ -58,6 +62,56 @@ describe('selectors', () => {
...
+      it('returns just empty ', () => {
```
```suggestion
      it('returns empty string', () => {
```

app/javascript/selectors/detailSelector.test.js
```diff
@@ -198,6 +254,117 @@ describe('selectors', () => {
...
+                  status: status,
+                },
+              },
+            },
+          },
+        };
+      };
+      it('return description when status is UNCONFIRMED', () => {
+        const state = getUserStatusDescription('UNCONFIRMED');
+        expect(userStatusDescription(state)).toEqual(
```
kelfink: What if we move the expect clause into this getUserStatusDescription method, so that we just call 
```
expectStatusDescriptionFor('UNCONFIRMED').toEqual('User has been created but not confirmed.');
```
I think it would be much more succinct.

app/javascript/selectors/detailSelector.test.js
```diff
@@ -198,6 +254,117 @@ describe('selectors', () => {
...
+                user: {
+                  status: status,
+                },
+              },
+            },
+          },
+        };
+      };
+      it('return userStatus friendly text when status is CONFIRMED ', () => {
+        const state = getUserStatus('CONFIRMED');
```
kelfink: same as userStatusDescription above.  Moving the expectation into the getter would make the tests more tidy

app/javascript/selectors/detailSelector.test.js
```diff
@@ -58,6 +62,56 @@ describe('selectors', () => {
...
+    });
+
+    it('return Inactive when enabled is false ', () => {
+      const state = getAccoutStatus(false);
+      expect(selectAccountStatus(state)).toEqual('Inactive');
+    });
+  });
+
+  describe('#selectStartDate', () => {
+    const getStartDate = startDate => {
```
RamuKammagani-osi: Suggestion: Instead of creating a separate function in each it block you can do something like creating a single function which returns state with different attributes and call that function in it block so that you can reduce at least 50 lines of code.

Sushmareddy-OSI:
retest this please

app/javascript/selectors/detailSelector.js
```diff
@@ -32,3 +70,40 @@ export const disableRolesDropDown = state =>
...
+    COMPROMISED: 'User is disabled due to a potential security threat.',
+    UNKNOWN: 'User status is not known.',
+    RESET_REQUIRED: 'Need to reset user.',
+    FORCE_CHANGE_PASSWORD: 'User has never logged in.',
+  };
+  return description[status] || '';
+};
+
+export const userStatus = state => {
+  const status = safeGet(state, 'fetchDetails.details.records.user.status');
```
Nasir2712: line 89 also has the same safeget, you sure this is correct?

Sushmareddy-OSI: yes, we need but I changed as generic method to get status

Nasir2712:
I understand code climate rule for keeping max lines of code per file as 250, is there any way we can turn it off for test files? we have so many scenarios that we try to cover not sure if we need the specific limit here.

Sushmareddy-OSI:
> I understand code climate rule for keeping max lines of code per file as 250, is there any way we can turn it off for test files? we have so many scenarios that we try to cover not sure if we need the specific limit here.

discussed with @kelfink on this, suggested not to turn off

sunilb0575:
The only way I feel better is to breakdown all our big components into smallest possible components, that way writing a

PR #220, opened 2018-11-01
COG-703-Fix on save error when roles dropdown is disabled

sunilb0575:
-https://osi-cwds.atlassian.net/browse/COG-703

-Fix for error, when roles dropdown is disabled.


app/javascript/_services/users/user.service.js
```diff
@@ -35,11 +35,12 @@ class UserService {
       });
   }
 
-  static saveUserDetails(id, details) {
+  static saveUserDetails(id, details, isRolesDisabled) {
+    const roles = isRolesDisabled ? undefined : details.roles;
```
ratneshraval: Is there a possibility where `details.roles` has data but `isRolesDisabled` is true.
Reason for asking: 
If when `isRolesDisabled` is true, the details.roles is always null, then we can avoid this if statement. 
If `isRolesDisabled` is simply overruling `details.role` then this is fine.

sunilb0575: Yes, there is a possibility for ```details.roles``` to have data and also ```isRolesDisabled``` to be true.

PR #223, opened 2018-11-07
CODE-52 implement standardized eslint rules

NagaPradeepGajula-osi:
- [Finalize linting and syncing with other digital services](https://osi-cwds.atlassian.net/browse/CODE-52)

- This PR involves configuring eslint/standard config and some of the best practices followed in react world. These were debated and finalized with the respective team leads.


kelfink:
Oh boy this is big.  Any way to separate this into a set of single changes, like 'remove all semicolons', and just to that as a single sweep ?

If I understand the changes here,  we might also be better off cleaning up the code climate changes first in a separate PR....  Is this PR changing the policy on number of returns in a function, or is the PR changing the actual number of returns in some of our code ?

NagaPradeepGajula-osi:
@kelfink I would like to keep it this way as every single acts like a single PR with specific changes. Removing semicolons into a separate PR would be the same as here with just one commit.

NagaPradeepGajula-osi:
And as per the rest of codeClimate issues being reported I haven't changed anything but remove semicolons. Do you guys have anything thoughts on fixing these too many returns?

NagaPradeepGajula-osi:
Also, this PR is for ESLINT rules only, whereas the limit on returns is reported by Code-climate default rules itself. I don't have any limit imposed on the returns in this PR at all.

kelfink:
Possibly the 'too many returns' is a legacy item we ignored before, or which became more restricted after a commit.  I'll look into that

NagaPradeepGajula-osi:
retest this please

NagaPradeepGajula-osi:
seems like the master merge didn't go well. Rebased and force pushed able to build locally.

app/javascript/_utils/codeToTextTranslator.js
```diff
@@ -0,0 +1,28 @@
+export function userStatusDescriptionTranslator(status) {
```
RamuKammagani-osi: Functions in these files are moved to detailSelector and this file is deleted, looks like your rebase with master din't go well here as well. Can you recheck once again?

app/javascript/_utils/codeToTextTranslator.test.js
```diff
@@ -0,0 +1,33 @@
+import { userStatusDescriptionTranslator, userStatusTranslator, accountStatus } from './codeToTextTranslator'
```
RamuKammagani-osi: same here

app/javascript/_utils/codeToTextTranslator.js
NagaPradeepGajula-osi: ok i will look into it.

NagaPradeepGajula-osi:
@RamuKammagani-osi nice catch on the deprecated files made changes and update the PR.

PR #228, opened 2018-11-09
COG-853 Add check to hide text when no value

RamuKammagani-osi:
- https://osi-cwds.atlassian.net/browse/COG-853

- Added extra check so that When no value for registration email resent field it wont display the text 'Registration email Resent'


app/javascript/views/userDetail/UserDetailEdit.jsx
```diff
@@ -89,18 +89,24 @@ const UserDetailEdit = ({
...
                     {userStatusDescription}
                     {details.status === 'FORCE_CHANGE_PASSWORD' && (
-                      <div className="resend-email-btn">
-                        <div className="resend-email-text">
-                          {`Registration email resent:`}
-                          <br />
-                          {checkDate(details.last_registration_resubmit_date_time)}
+                      <div>
+                        <div>
+                          {details.last_registration_resubmit_date_time && (
```
Nasir2712: why is this if else nested here? and in UserDetailShow it has a null check? can you refactor this in both places to look the same?

RamuKammagani-osi: I did that because the resend button and resubmitted data and text should be separate. But in show card there is no button so i just added a check to the same condition.

app/javascript/views/userDetail/UserDetailShow.jsx
```diff
@@ -69,13 +69,14 @@ const UserDetailShow = ({
...
                     {userStatusDescription}
-                    {details.status === 'FORCE_CHANGE_PASSWORD' && (
-                      <div className="resend-email-text">
-                        {`Registration email resent:`}
-                        <br />
-                        {checkDate(details.last_registration_resubmit_date_time)}
-                      </div>
-                    )}
+                    {details.status === 'FORCE_CHANGE_PASSWORD' &&
+                      details.last_registration_resubmit_date_time !== null && (
```
Nasir2712: you could just do ```details.last_registration_resubmit_date_time &&``` instead of doing !==null.

PR #229, opened 2018-11-10
Cog 857 stable user deactivation in tests

kelfink:
- https://osi-cwds.atlassian.net/browse/COG-857

-  Fixes issues with deactivating previously-added users on the last page of the user list. Should be faster and ore reliable now.


kelfink:
retest this please

PR #232, opened 2018-11-13
Fix the display of two alerts after edit when user is added

RamuKammagani-osi:
- n/a

- This fixes the issue of displaying two alerts at the same time, After adding a user and if  user is  edited and when save is clicked it displays two alerts.


kelfink:
retest this please

PR #233, opened 2018-11-14
Cog 857 wait after add

kelfink:
- COG-857

-  Latest acceptance test is unstable because, it looks like, the user has not yet been retrieved from the IDM source for display.  Wait until the


spec/acceptance/add_user_spec.rb
```diff
@@ -31,6 +31,9 @@
     expect(page).to have_button('Add User')
 
     click_button 'Add User'
+    page.find('div.success-message')
+    expect(find(:label, 'CWS Login').find(:xpath, './/../span').text).to eq valid_racfid
+
     expect(page).to have_button('Edit')
```
RamuKammagani-osi: i think  after the success message when user is added we should test whether the page has edit button or not.

  expect(page.find('div.success-message').text).to eq(message)
  page.evaluate_script('window.location.reload()')
  expect(page).to have_button('Edit')

I think this sequence would be nice after the success message, reload the page and then looking for the edit button.

kelfink: See updates

PR #235, opened 2018-11-15
COG-640 Update registration date& time  after editing a user

RamuKammagani-osi:
- https://osi-cwds.atlassian.net/browse/COG-640

- Updates  registration emails sent date and time after editing a user, Updates when user is comes out of edit mode.


kelfink:
How do these changes relate to the title, regarding " date& time after editing a user"

RamuKammagani-osi:
> How do these changes relate to the title, regarding " date& time after editing a user"

After the user clicked on resend button the date & time saved in db, So i'm making a fetch call to get the updated details when user is clicked on save or cancel button.

app/javascript/views/userDetail/UserDetail.jsx
```diff
@@ -29,6 +29,7 @@ export default class UserDetail extends Component {
   onSaveDetails = () => {
     const { details, isRolesDisabled, match, actions } = this.props
     actions.saveUserDetailsActions(match.params.id, details, isRolesDisabled)
+    this.props.actions.fetchDetailsActions(this.props.match.params.id)
```
Nasir2712: when you do not hit the registration button and just hit save, this line will fetch all the old details, it should be removed.

PR #236, opened 2018-11-15
COG-640 Fix details update

RamuKammagani-osi:
-  https://osi-cwds.atlassian.net/browse/COG-640

- Fixes the details update and updates date & time for registration email resent.


app/javascript/reducers/detailsReducers.js
```diff
@@ -54,6 +57,44 @@ function fetchDetails(
...
+            user: {
+              ...state.details.records.user,
+              last_registration_resubmit_date_time: dateTime,
+            },
+          },
+        },
+        error: null,
+      }
+
+    case actionTypes.RESEND_REGISTRATION_EMAIL_API_CALL_FAILURE:
```
kelfink: With the addition of these cases, code climate is complaining.  Let's add move these to a new reduce method (same class for now is OK) and add it to the exports, then add to reducers/index.js combineReducers call.  Others reducers can be extracted as well.

app/javascript/reducers/detailsReducers.js
```diff
@@ -54,6 +57,44 @@ function fetchDetails(
...
+      }
+
+      const date = DateTime.local().toSQL({ includeOffset: false })
+      const dateTime = date.split('.')[0]
+
+      return {
+        ...state,
+        fetching: false,
+        resendEmailStatus: status.response,
+        // resendEmailUserId: state.resendEmailUserId.concat(action.id),
```
Nasir2712: we still need this please uncomment this line.

app/javascript/reducers/detailsReducers.js
Nasir2712: these reducers cannot be extracted because of details object. the resend registration email was moved to this reducer because the registration date is nested inside the user detail object, moving this out of fetchDetails reducer would not allow accessing the details state.

app/javascript/reducers/detailsReducers.js
RamuKammagani-osi: Updated

PR #240, opened 2018-11-19
Update to cwds-ruby-1.1 base image

kelfink:
- 
https://osi-cwds.atlassian.net/browse/COG-209

-
Use the new rails 1.1 base image to begin docker build of CAP container for increased security.


ratneshraval:
Let's verify if the base is issue free before we switch over.

PR #241, opened 2018-11-19
COG-919 Add Inactive Status Checkbox

Nasir2712:
- https://osi-cwds.atlassian.net/browse/COG-919

- add include inactive checkbox and includeInactive attribute to search wherever it needs to be fired.
- rails es part is not done in this PR.


app/javascript/views/usersList/UsersList.jsx
```diff
@@ -36,6 +38,7 @@ class UserList extends PureComponent {
       this.props.actions.setSearch([
         { field: 'last_name', value: this.props.lastName },
         { field: 'office_ids', value: this.props.officeNames },
+        { field: 'enabled', value: !this.props.includeInactive },
```
kelfink: My interpretation of this is "enabled:  false". or "enabled: true".  are being sent always.  So you're asking either for "active" or "inactive" users.  If you want elastic search to include both true and false, don't send this parameter at all

Nasir2712: ok

app/javascript/views/usersList/UsersList.jsx
```diff
@@ -44,6 +47,15 @@ class UserList extends PureComponent {
     this.setState({ addUser: true })
   }
 
+  handleCheckBoxChange = () => {
+    this.props.actions.handleCheckBoxChangeActions()
+    this.props.actions.setSearch([
+      { field: 'last_name', value: this.props.lastName },
+      { field: 'office_ids', value: this.props.officeNames },
+      { field: 'enabled', value: this.props.includeInactive },
```
kelfink: As above we should not send 'enabled' if we don't want to filter it.  It would be safe to send an empty string, I believe.  Similar to office_ids, if it's empty the rails layer ignores it.

Nasir2712: good point, i'll update this

Nasir2712: I may have to add some conditions to when to send the enabled parameter.

app/javascript/views/usersList/UsersList.jsx
Nasir2712: same here, i need conditions to check when to send the enabled param or not.

app/javascript/_services/users/user.service.js
```diff
@@ -6,6 +6,9 @@ class UserService {
   }
 
   static search({ query = [], sort = [], size = 10, from = 0 }) {
+    if (query.length > 0) {
+      query[2].value = query[2].value.toString()
```
kelfink: what's the magic number 2 all about?

kelfink:
```suggestion
      const enabledQuery = query.find(obj => {
        return obj.field === "enabled";
      });
      enabledQuery.value = enabledQuery.value.toString();
```

app/javascript/_services/users/user.service.test.js
```diff
@@ -30,6 +30,23 @@ describe('UserService', () => {
...
+        },
+        {
+          field: 'office_ids',
+          value: ['north', 'south', 'east', 'west'],
+        },
+        { field: 'enabled', value: true },
+      ]
+
+      UserService.search({ query })
+      expect(query[2].value).toEqual('true')
```
kelfink:
```suggestion
      const enabledQuery = query.find(obj => {
        return obj.field === 'enabled';
      });
      expect(enabledQuery.value).toEqual('true')
```

app/javascript/_services/users/user.service.js
Nasir2712: interestingly I did the same thing without looking at the suggestion lol

app/javascript/views/usersList/UsersList.jsx
Nasir2712: done

PR #242, opened 2018-11-20
Add elastic search filtering on enabled users instead of all

kelfink:
- https://osi-cwds.atlassian.net/browse/COG-920

- Adds a 'enabled' as a possible parameter to the react-rails contract for searching for users.  If enabled is provided, its value will be compared against the same value in Elastic search user documents.


lib/elastic/user_query_builder.rb
```diff
@@ -31,16 +31,18 @@ def self.query_leaves(query)
       end.compact
     end
 
+    SUBQUERIES = {
+      office_ids: ->(value) { { terms: { 'office_id.keyword': value } } unless value.empty? },
+      last_name: ->(value) { { match_phrase_prefix: { last_name: value } } unless value.empty? },
+      enabled: ->(value) { { terms: { enabled: value } } unless value.empty? }
```
Nasir2712: value.empty? would throw an error if we send enabled value as boolean when we have to show only active users.

Nasir2712: also, ```terms``` in the elastic search query is throwing an error, either match or match_phrase_prefix are working work

kelfink: Ah, you mean if value is true (boolean value) then it blows up with "undefined method `empty?' for true:TrueClass". Gotcha.  I'll make a fix to protect it.

kelfink: Only in the case where a boolean is there instead of a string, you mean?

Nasir2712:
```{"query":{"bool":{"must":[{"terms":{"enabled":"true"}}]}},"from":0,"size":50,"sort":[{"last_name.for_sort":{"order":"asc"}},{"first_name.for_sort":{"order":"asc"}}]}
``` 
```{"query":{"bool":{"must":[{"terms":{"enabled":true}}]}},"from":0,"size":50,"sort":[{"last_name.for_sort":{"order":"asc"}},{"first_name.for_sort":{"order":"asc"}}]}
``` 
both queries throw errors in dora, i think the keyword ```terms``` only work for array of values and not string or boolean.

Nasir2712: I had this comment when I was sending a boolean, but now after the new updates, I think this would be fine unless you want more protection.

kelfink: Good catch.  Actually it was meant to be "term" instead of "terms"

lib/elastic/user_query_builder.rb
```diff
@@ -31,16 +31,18 @@ def self.query_leaves(query)
       end.compact
     end
 
+    SUBQUERIES = {
+      office_ids: ->(value) { { terms: { 'office_id.keyword': value } } unless value.empty? },
+      last_name: ->(value) { { match_phrase_prefix: { last_name: value } } unless value.empty? },
+      enabled: ->(value) { { term: { enabled: value.to_s } } unless value.to_s.empty? }
```
Nasir2712: i think it's fine if you just value, you dont have to convert this to a string again.
```if (query.length > 0) {
      const enabledField = query.find(q => q.field === 'enabled')
      enabledField.value = enabledField.value.toString()
    }
```
this part right here is already converting the value to string.

PR #247, opened 2018-11-27
COG-922 Display DateTime when resend button is clicked in edit/show card

RamuKammagani-osi:
- https://osi-cwds.atlassian.net/browse/COG-922

- Displays the response DateTime in UI Show/Edit cards from API when the resend button is clicked. 
- Modified according to the changes in the contract from API.


lib/users/user_repository.rb
```diff
@@ -55,8 +55,12 @@ def add_user(parameters, token)
     end
 
     def resend_registration_email(parameters, token)
-      response = @http_service.get('/perry/idm/users/resend', parameters, token)
-      response.status
+      response = @http_service.post(
+        "/perry/idm/users/#{parameters['id']}/registration-request",
```
kelfink: If parameters is still being passed into this method, why not send it to the http_service?  I think setting a default value for parameters (second of three params) in the post method could be problematic.

lib/infrastructure/http_service.rb
```diff
@@ -17,7 +17,7 @@ def get(url, parameters = {}, token)
       return Faraday::Response.new(status: 404)
     end
 
-    def post(url, parameters, token)
+    def post(url, parameters = {}, token)
```
kelfink: I've never tried defaulting a middle parameter and not further ones.  Seems weird.

lib/users/user_repository.rb
RamuKammagani-osi: Well, if no second parameter was passed to that method it will take the default value as empty, I Still dint get it why is it problematic, But anyway updated it the latest commit.

lib/infrastructure/http_service.rb
RamuKammagani-osi: Updated in the latest commit.

PR #250, opened 2018-11-28
Set timezone to Los Angeles for acceptance testing

kelfink:
- COG-922

-  An acceptance test is failing due to mis-match between timezones.  We need to inform the container what timezone to operate in.


kelfink:
retest this please

PR #251, opened 2018-11-28
Fix console error when button is clicked & hover on button

RamuKammagani-osi:
- https://osi-cwds.atlassian.net/browse/COG-912

-  Hover reflects a little and changes the color a bit


Nasir2712:
I tried this locally, but the hover CSS background-color: none is an invalid value, can you check?

RamuKammagani-osi:
> I tried this locally, but the hover CSS background-color: none is an invalid value, can you check?

RamuKammagani-osi:
> I tried this locally, but the hover CSS background-color: none is an invalid value, can you check?

Updated

app/javascript/views/userDetail/UserDetailShow.jsx
```diff
@@ -96,7 +96,7 @@ const UserDetailShow = ({
                             btnClassName="primary"
                             btnName="Resend Invite"
                             onClick={onResendInvite}
-                            disabled={disableResendEmailButton | disableEditBtn}
+                            disabled={disableEditBtn || disableResendEmailButton}
```
Nasir2712: not sure why disable edit button prop is needed here?

RamuKammagani-osi: Yes it works, But when the user is not editable resend button should be disabled, so i added  i added that

Nasir2712: ok, I thought we hid the button, in that kind of scenario, anyways thanks for the clarification.

RamuKammagani-osi: Yes, we are going to hide the button in the nearest future.

PR #253, opened 2018-11-29
Set TZ value directly on test container

kelfink:
Repair timezone settings by setting the timezone directly in the docker-compose.yml


kelfink:
Added another commit based on discussions with other devs.  JenksinfileAcceptance needs to pass in the value from the Jenkins config vars, in addition to the docker-compose.yml configuration

PR #255, opened 2018-12-01
Wait for loading/result message for better load results in acceptance…

kelfink:
… tests

-  COG-815 

-Improvement on the process of waiting for verification details to load when adding a user.


spec/acceptance/add_user_spec.rb
```diff
@@ -49,7 +49,7 @@
 
     click_on('Resend Invite')
 
-    date_time = Time.now.strftime('%B %d, %Y %I:%M %p')
+    date_time = Time.now.strftime('%B %-d, %Y %I:%M %p')
```
kelfink: No leading 0 should be expected.  %d by default pads with a 0.  🤷‍♂️

PR #256, opened 2018-12-03
COG-707 Card Header Value change based on role

RamuKammagani-osi:
- https://osi-cwds.atlassian.net/browse/COG-707

- Card Header Text changes based on the user role.


app/javascript/reducers/userListReducers.js
```diff
@@ -24,9 +24,7 @@ const initialValue = {
   fetching: false,
   error: null,
   inputData: {},
-  countyName: '',
-  userFirstName: '',
-  userLastName: '',
+  userAccountDetails: {},
```
kelfink: What about naming this "adminAccountDetails". so we can more easily distinguish this variable from ones that refer to the system users, as opposed to the admin user who is logged in ?

RamuKammagani-osi: I like it! adminAccountDetails makes more sense than userAccountDetails, Updated in latest commits

PR #259, opened 2018-12-06
COG-510 Cognito - Add acceptance tests for login screen

Nasir2712:
- https://osi-cwds.atlassian.net/browse/COG-510

- Since the MFA code tests have already been written, I wrote acceptance tests for forgot password page scenarios.


spec/acceptance/support/login_helper.rb
```diff
@@ -35,6 +35,17 @@ def cognito_login
...
   end
 
+  def click_forgot_password_link
+    visit ENV.fetch('RAILS_RELATIVE_URL_ROOT', '/')
+    return unless current_url.include?('login')
+    click_link 'Forgot your password?'
+  end
+
+  def reset_password
+    fill_in 'Email', with: ENV.fetch('COGNITO_USERNAME', 'y_test111+role1@outlook.com')
```
kelfink: not sure we should have a default email address that looks real.  Or is this real?

Nasir2712:
sure

spec/acceptance/support/login_helper.rb
Nasir2712: if you have any automated email that works when email submit, i'll be happy to use that. The ones that we are using don't work for reset password.

Nasir2712:
I'll add those test cases as well. I'll move this to WIP.

PR #260, opened 2018-12-06
Cog 951 edit and save user email

Sushmareddy-OSI:
[COG-933](https://osi-cwds.atlassian.net/browse/COG-933)
[COG-951](https://osi-cwds.atlassian.net/browse/COG-951)

- Able to edit and save email.


Gemfile.lock
```diff
@@ -368,7 +368,7 @@ DEPENDENCIES
   byebug
   capybara (~> 2.13)
   capybara-screenshot
-  chromedriver-helper (~> 1.2.0)
+  chromedriver-helper
```
kelfink: why?  Did we need a newer version?

Sushmareddy-OSI:
To fix CodeClimate issues, need refactoring on details.jsx and detailsReducer.jsx files. Will address in another PR. Since the functionality is in place it will be ready for the testers.

app/javascript/_services/users/user.service.test.js
```diff
@@ -104,19 +105,22 @@ describe('UserService', () => {
...
+      it('calls #patch ApiService when permission, email & roles are empty', () => {
         body.permissions = []
         body.roles = ''
+        body.email = ''
         UserService.saveUserDetails(id, {
           enabled: true,
           permissions: [],
           roles: '',
           first_name: 'firstName',
+          email: '',
```
kelfink: Isn't email required?  Why would we save if we know it's invalid?

app/javascript/actions/detailActions.js
```diff
@@ -5,13 +5,19 @@ import {
...
   CLEAR_SAVE_ALERT,
+  HANDLE_EMAIL_INPUT_CHANGE,
 } from './actionTypes'
 
 export const handleDropdownChangeAction = (name, value) => ({
   type: HANDLE_DROPDOWN_CHANGE,
   payload: { name, value },
 })
 
+export const handleInputChangeAction = value => ({
```
kelfink: does this method have the right name?  Looks like it's only for the email input.

Gemfile.lock
Sushmareddy-OSI: It forced me to do `bundle install`, with that some of the gems got updated!

app/javascript/reducers/detailsReducers.js
```diff
@@ -104,6 +107,30 @@ function fetchDetails(
       }
     }
 
+    case actionTypes.HANDLE_EMAIL_INPUT_CHANGE: {
+      const email = action.payload.value
+      const validateEmail = /^[a-zA-Z0-9_!#$%&’*+/=?`'{^.-]*@(?:[A-Z0-9-]+\.)+[A-Z]{2,6}$/i.test(email)
```
kelfink: let's pull this line out to a method 'validateEmail' that takes the email string and returns boolean

app/javascript/reducers/detailsReducers.js
```diff
@@ -104,6 +107,30 @@ function fetchDetails(
...
+      validateEmail ? (state.emailErrorMessage = '') : (state.emailErrorMessage = 'Please enter a valid email.')
+      state.emailErrorMessage === '' ? (state.disableActionBtn = false) : (state.disableActionBtn = true)
```
kelfink: for 113 and 114, it seems tangled, as yore using 113's result to perform the logic on 114, but they really are the same case.
 I suggest direct logic to make it more clear...
```suggestion
      if (validateEmail(email) {
        state.emailErrorMessage = ''
        state.disableActionBtn = false
      } else {
         state.emailErrorMessage = 'Please enter a valid email.'
         state.disableActionBtn = true
       }
```
Or, consider putting the emailErrorMessage directly on the component render, but hiding that when disableActionBtn state is true.  Might eliminate this code altogether.

app/javascript/reducers/detailsReducers.test.js
```diff
@@ -73,6 +73,58 @@ describe('reducer', () => {
...
+      }
+      const state = { details: details }
+      expect(fetchDetails(state, handleInputAction)).toEqual({
+        details: {
+          records: { user: { email: value } },
+        },
+        displayAlert: false,
+        disableActionBtn: true,
+        isEmailValid: true,
+        emailErrorMessage: 'Please enter a valid email.',
```
kelfink: same text.  More evidence it would be better to put this text directly into, perhaps, a span, which we render only if isEmailValid is false.

app/javascript/reducers/detailsReducers.test.js
```diff
@@ -111,6 +164,8 @@ describe('reducer', () => {
       disableActionBtn: false,
       resendEmailUserId: [],
       resendEmailStatus: null,
+      emailErrorMessage: '',
```
kelfink: if you pull the email error message out and leave it just in the component, these would go away and not need to be tested.

app/javascript/views/addUser/VerifyUser.jsx
```diff
@@ -1,7 +1,7 @@
 import React from 'react'
 import PropTypes from 'prop-types'
 import Cards from '../../common/Card'
-import { InputComponent } from 'react-wood-duck/'
+import { InputComponent } from 'react-wood-duck'
```
kelfink: ;-)

app/javascript/views/userDetail/UserDetailEdit.jsx
```diff
@@ -67,7 +67,17 @@ const UserDetailEdit = ({
...
+              <InputComponent
+                id="InputEmail"
+                label="Email"
+                fieldClassName="form-group"
+                type="email"
+                placeholder="Add Email Address"
+                value={details.email}
+                onChange={event => onInputChange(event.target.value)}
+                validationError={isEmailValid}
+                validationErrorMessage={emailErrorMessage}
```
kelfink: So if we only have the one message, seems like we can just hard-code it here and not pass it around, also avoiding tests.

app/javascript/actions/detailActions.js
Sushmareddy-OSI: given a generic name so that we can also use the same method for other input fields if necessary in future

Gemfile.lock
kelfink: /shrug.  ok

app/javascript/actions/detailActions.js
kelfink: But the enclosed type is not generic.  So it's not consistent with how handleDropDownChangeAction works.

spec/acceptance/user_edit_details_spec.rb
```diff
@@ -114,28 +114,36 @@
...
+    original_email_address = detail_page_value('Email')
 
     click_on('Edit')
-    changed_new_status = change_status(new_status)
+    expect(page).to have_button('save', disabled: true)
+    expect(page).to have_button('Cancel', disabled: false)
+    fill_in('Email', with: 'cwds3raval', match: :prefer_exact)
+    # bad email address won't let us proceed
+    expect(page).to have_button('save', disabled: true)
+    expect(page).to have_content('Please enter a valid email.')
```
Nasir2712: this should be 'Please enter the valid email.'
screen shot 2018-12-07 at 4 33 33 pm

app/javascript/reducers/detailsReducers.js
Nasir2712: yea, i agree what I would suggest is to move this validation part out to the selector as well to keep the reducer logic clean.

app/javascript/views/userDetail/UserDetailEdit.jsx
```diff
@@ -67,7 +66,17 @@ const UserDetailEdit = ({
...
+              <InputComponent
+                id="InputEmail"
+                label="Email"
+                fieldClassName="form-group"
+                type="email"
+                placeholder="Add Email Address"
+                value={details.email}
+                onChange={event => onInputChange(event.target.value)}
+                validationError={!isEmailValid}
+                validationErrorMessage={'Please enter the valid email'}
```
Nasir2712: I think this needs to be changed to 'Please enter a valid email.'

spec/acceptance/user_edit_details_spec.rb
Sushmareddy-OSI: I am using the same error message as on verify user page. It says, 'Please enter a valid email"

app/javascript/views/userDetail/UserDetailEdit.jsx
Sushmareddy-OSI: Agree!!

app/javascript/selectors/detailSelector.test.js
```diff
@@ -369,6 +375,40 @@ describe('selectors', () => {
...
+      const state = getState({ email: 'hello@', disableActionBtn: false })
+      expect(disableActionButton(state)).toEqual(true)
+    })
+
+    it('return false when email is valid and disableActionBtn is false', () => {
+      const state = getState({ email: 'hello@gmail.com', disableActionBtn: false })
+      expect(disableActionButton(state)).toEqual(false)
+    })
+
+    it('return true when email is not valid and disableActionBtn is true', () => {
```
kelfink: You could remove his and simplify the it block on line 396 to just :
```suggestion
  it('return true when email is not valid', () => {
       const state = getState({ email: 'hello@', disableActionBtn: false })
       expect(disableActionButton(state)).toEqual(true)
     })
```

kelfink: because overall any invalid email must cause us to return true

PR #264, opened 2018-12-12
Fix failing cognito acceptance tests

Nasir2712:
- https://osi-cwds.atlassian.net/browse/COG-510

-

### DIFF
--- a/spec/acceptance/support/login_helper.rb
+++ b/spec/acceptance/support/login_helper.rb
@@ -66,6 +66,10 @@ module LoginHelper
     click_button 'Reset my password'
   end
 
+  def logout_link
+    visit "#{ENV.fetch('COUNTY_ADMIN_WEB_BASE_URL', '/')}/logout"
+  end
+
   private
 
   def verify_account
--- a/spec/acceptance/support/login_helper.rb
+++ b/spec/acceptance/support/login_helper.rb
@@ -48,8 +48,8 @@ module LoginHelper
     visit ENV.fetch('RAILS_RELATIVE_URL_ROOT', '/')
     return unless current_url.include?('login')
 
-    fill_in 'Email', with: ENV.fetch('COGNITO_USERNAME', 'no-reply@osi.ca.gov')
-    fill_in 'Password', with: ENV.fetch('COGNITO_PASSWORD', 'password')
+    fill_in 'Email', with: 'y_test111+role2@outlook.com'
+    fill_in 'Password', with: 'Password123*'
     click_on 'Sign In'
     invalid_mfa
   end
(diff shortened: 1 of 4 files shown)


spec/acceptance/support/login_helper.rb
```diff
@@ -48,8 +48,8 @@ def cognito_login_with_invalid_mfa
     visit ENV.fetch('RAILS_RELATIVE_URL_ROOT', '/')
     return unless current_url.include?('login')
 
-    fill_in 'Email', with: ENV.fetch('COGNITO_USERNAME', 'no-reply@osi.ca.gov')
-    fill_in 'Password', with: ENV.fetch('COGNITO_PASSWORD', 'password')
+    fill_in 'Email', with: 'y_test111+role2@outlook.com'
```
kelfink: what?  no?  why?

Nasir2712: what happened?

spec/acceptance/user_forgot_password_spec.rb
```diff
@@ -5,6 +5,9 @@
 require 'axe/rspec'
 
 feature 'User Forgot Password Page' do
+  before(:all) do
+    visit "#{ENV.fetch('COUNTY_ADMIN_WEB_BASE_URL', '/')}/logout"
```
kelfink: I suggest putting this in the login_helper. (maybe rename it if that bothers you) so that it's reusable on other specs.
Also, I had (locally) implemented that in a more UI-driven way, popping open the avatar link and clicking the 'logout' piece.  But this is more efficient, for wider use.

Nasir2712: cool

spec/acceptance/user_login_spec.rb
```diff
@@ -5,6 +5,9 @@
 require 'axe/rspec'
 
 feature 'User Sign in' do
+  before(:all) do
+    visit "#{ENV.fetch('COUNTY_ADMIN_WEB_BASE_URL', '/')}/logout"
```
kelfink: same re-use comment.  good efficient way to do it though

spec/acceptance/support/login_helper.rb
Nasir2712: so this method is only being used in one test case, I need the credentials changed for this method because once we log out and try to log in again with the same credentials the MFA page does not pop up. In order for me to access this MFA page again, I need to change the credentials.

spec/acceptance/user_login_spec.rb
Nasir2712: done

spec/acceptance/support/login_helper.rb
kelfink: Ah, this is a failure method.  I get it now.

PR #267, opened 2018-12-18
Cog 473 avoid redirect to logout

kelfink:
- https://osi-cwds.atlassian.net/browse/COG-473
Force user to home page instead of logout URL's
-  Force user to the home page instead of the logout URL when authenticating, otherwise they may experience double login issue as described in COG-473


spec/lib/infrastructure/cwds_authenticator_spec.rb
```diff
@@ -28,6 +28,19 @@ module Infrastructure
...
+
+        context('when the url ends with logout') do
+          let(:environment_bad) { Rack::MockRequest.env_for('http://example.com/cap/logout', {}) }
+          it 'Does not provide the logout path in the callback' do
+            Feature.run_with_activated(:authentication) do
+              allow(security_policy).to receive(:validate_access)
+                .with(instance_of(Rack::Request)).and_return(nil)
+              status, headers = cwds_authenticator.call(environment_bad)
+              expect(status).to eq 301
+              expect(headers['Location']).to eq 'https://perry.test/authn/login?callback=http://example.com'
```
ratneshraval: I think this behavior should be that if we have `http://example.com/cap/logout` as callback, the output is `http://example.com/cap`.
So only `/logout` is removed.

spec/lib/infrastructure/cwds_authenticator_spec.rb
```diff
@@ -28,6 +28,19 @@ module Infrastructure
             expect(headers['Location']).to eq 'https://perry.test/authn/login?callback=http://example.com/'
           end
         end
+
+        context('when the url ends with logout') do
```
Nasir2712: can we add one more scenario for this test case if the uri does not have '/logout'?

PR #268, opened 2018-12-18
Cog 608 display proper message when user is not found

Sushmareddy-OSI:
[COG-608](https://osi-cwds.atlassian.net/browse/COG-608)

For county admins trying to access user profile of users in another county, proper message should be displayed instead of showing a blank page


app/javascript/views/userDetail/UserDetail.jsx
```diff
@@ -114,116 +114,95 @@ export default class UserDetail extends Component {
...
-            disableEditBtn,
-            accountStatus,
-            assignedPermissions,
-            startDate,
-            userStatusDescription,
-            userStatus,
-            officeName,
-            resentRegistrationEmailDateTime,
-            isEmailValid
+          {this.props.fetchDetailsError ? (
```
Nasir2712: ErrorMessage is a component we have in common folder that we are using I guess to show error messages. If you just update the alert inside that component to the newest one, I think every error message will be updated.

app/javascript/views/userDetail/UserDetail.jsx
```diff
@@ -114,116 +114,95 @@ export default class UserDetail extends Component {
     resentRegistrationEmailDateTime,
     isEmailValid
   ) => {
-    return XHRStatus !== 'ready' ? (
-      <Cards>
-        <span>{'Loading...'}</span>
-      </Cards>
-    ) : (
+    return details && details.id ? (
```
Nasir2712: if code climate is an issue for this, I think we can move this.renderCards code to a separate file so that it will be much easier to control both lines and logic.

lib/infrastructure/http_service.rb
```diff
@@ -13,8 +12,6 @@ def initialize(base_url = Rails.configuration.micro_services['perry_api_base_url
 
     def get(url, parameters = {}, token)
       http_connection.get("#{url}?#{parameters.to_query}&token=#{token}")
-    rescue StandardError
```
Nasir2712: let's keep it as it is, since we are handling 401's in the middleware this shouldn't be an issue to us.

app/javascript/selectors/detailSelector.js
```diff
@@ -75,6 +75,12 @@ export const disableActionButton = state => {
   return safeGet(state, 'fetchDetails.disableActionBtn')
 }
 
+export const fetchDetailsError = state => {
```
Nasir2712: I think if we use the error message component, we wouldn't need this selector.

app/javascript/views/userDetail/UserDetail.jsx
Sushmareddy-OSI: The ErrorMessage Component is taking {error.user_message}, what if there is no user_message field in the error.?

lib/infrastructure/http_service.rb
Sushmareddy-OSI: okay!

app/javascript/views/userDetail/UserDetail.jsx
Nasir2712: if there is no user_message, the default props will kick in but it won't crash the page or anything like that. If we get the error object and don't have a user_message then that's also a backend issue. Also, feel free to modify the component to your liking if you think something is not right.

app/javascript/_services/users/user.service.test.js
```diff
@@ -172,6 +164,44 @@ describe('UserService', () => {
...
+        const error = 'error'
+        const id = 'someID'
+        getSpy2.mockReturnValue(Promise.reject(error))
+        UserService.fetchUserDetails(id)
+        expect(getSpy2).toHaveBeenCalledWith('/user_detail/someID')
+      })
+    })
+
+    it('calls #fetchUsersDetails ApiService', () => {
+      getSpy.mockReturnValue(Promise.resolve({}))
```
Nasir2712: were you trying to do something here? I don't see any expect statement here.

app/javascript/views/userDetail/UserDetail.test.js
```diff
@@ -212,6 +209,13 @@ describe('UserDetail', () => {
         expect(wrapper.find('UserDetailShow').length).toBe(1)
       })
 
+      it('should display <Alert/>', () => {
+        wrapper.setProps({
+          fetchDetailsError: 'Unauthorized',
```
Nasir2712: shouldn't this be a object?

app/javascript/_services/users/user.service.test.js
Sushmareddy-OSI: good catch! I was, but I added on the top. Thanks

PR #269, opened 2018-12-20
Display alerts with using message component

Sushmareddy-OSI:
-NA

- UserMessage Component to display the alerts.


Nasir2712:
you're right, the alert's are failing acceptance tests, needs to be fixed.

PR #270, opened 2018-12-21
COG-994 handle logic in selector and make UI components dumb in user detail page

RamuKammagani-osi:
- https://osi-cwds.atlassian.net/browse/COG-994

- Make UI components dumb in user detail page and handle logic in selectors.


app/javascript/containers/addUserContainer.js
```diff
@@ -17,6 +17,7 @@ function mapStateToProps(state) {
     dashboardUrl: '/',
     validateNewUserError: state.validateNewUser.error,
     officeName: officeName(state),
+    officePhoneNumber: formattedPhoneNumber(state),
```
Nasir2712: why not use a selector for this as well?

app/javascript/selectors/detailSelector.js
```diff
@@ -142,3 +142,22 @@ export const selectModifiedDetails = state => {
...
+export const lastLogin = state => safeGet(state, 'fetchDetails.details.records.user.last_login_date_time', '')
+
+export const resentRegistrationDate = state =>
+  safeGet(state, 'fetchDetails.details.records.user.last_registration_resubmit_date_time', '')
+
+export const formattedDateTime = date => {
+  return checkDate(date)
+}
+
+export const assignedRoles = state => {
```
Nasir2712: I feel like phone number, date, time need formatters because they need to be presented differently but assigned roles logic can be moved in the selector. As far as my understanding it does not need formatting. What do you think?

Sushmareddy-OSI: No, we need formatter for assignedRoles as well because we are getting assignedRole as 'county-admin', we need to match this with RolesList and return as label 'County Administrator'

Nasir2712: so that formatter logic cannot be moved to the selector?

RamuKammagani-osi: Formatters logic  for assignedRoles are being used in userList as well, so moving the logic to individual selector would create a confusion and as application grows its hard to debug, so that's the reason i created a selector which uses generic method to return label which can be used  for either userlist and user detail page or both or wherever it needed.

app/javascript/containers/addUserContainer.js
RamuKammagani-osi: yeah i'm using from selectors...validateNewUserSelector, Can you please elaborate whats your expectation here?

app/javascript/containers/detailsContainer.js
```diff
@@ -54,14 +59,18 @@ function mapStateToProps(state) {
     userListUrl: process.env.RAILS_RELATIVE_URL_ROOT ? process.env.RAILS_RELATIVE_URL_ROOT : '/',
     userDetailError: state.fetchDetails.saveDetailsError,
     fetchDetailsError: state.fetchDetails.fetchDetailsError,
-    resentRegistrationEmailDateTime: selectResendEmailDateTime(state),
+    resentRegistrationNewDateTime: formattedDateTime(resentRegistrationNewDateTime(state)),
```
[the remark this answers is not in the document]

app/javascript/containers/addUserContainer.js
Nasir2712: oh! so the name of the selector is formattedPhoneNumber, I thought it's a formatter, what do you think about the renaming the method name to match what we are using for selectors?

app/javascript/selectors/detailSelector.js
Nasir2712: I know I am kinda dragging this but I do have a suggestion for the future. 
Since we have a lot of common methods that are being used in the selector, I think we should create a new file called commonSelectors and move these methods to that file so that they can be exported to wherever they are needed. This way we can avoid confusion.
I think since we are using selectors and wrapping formatters inside a  selector we are adding an extra layer of effort for debugging.

PR #277, opened 2018-12-26
COG-1064-Add super-admin to roles

sunilb0575:
-https://osi-cwds.atlassian.net/browse/COG-1064

-Add Super-admin to roles


sunilb0575:
jenkins retest this please

app/javascript/selectors/userListSelector.js
```diff
@@ -36,6 +36,11 @@ export const cardHeaderText = state => {
   const roles = safeGet(state, 'userList.adminAccountDetails.roles', [])
   const countyName = safeGet(state, 'userList.adminAccountDetails.county_name', '')
   const role = roles || []
-  const cardHeaderText = role.includes('State-admin') ? 'State Administrator View' : `County: ${countyName}`
-  return cardHeaderText
+  if (role.includes('Super-admin')) {
+    return 'Global Administrator view'
```
RamuKammagani-osi: is this right?? As per my understanding there is no UI Change at all, is it confirmed with PO? just checking.

Sushmareddy-OSI:
Need to fix the codeClimate issue. Other than that, looks good!

app/javascript/_utils/checkAdminRoles.test.js
```diff
@@ -9,6 +9,14 @@ describe('#getAdminOfficeIDs', () => {
     expect(getAdminOfficeIDs(account)).toEqual(['1234'])
   })
 
+  it('returns empty array when roles have Super-admin ', () => {
```
RamuKammagani-osi: Can you also add 2 more scenarios when roles are null or undefined? I see you tested when account is undefined, But what if account exists and roles are undefined or null or some other text?

app/javascript/selectors/userListSelector.js
sunilb0575: Yes, It is confirmed from PO.

sunilb0575:
Sure. @RamuKammagani-osi @Sushmareddy-OSI I will create another PR for that code climate issue.

PR #280, opened 2019-01-03
COG-1078 Display worker phone number in user detail view/edit & verify user screens

RamuKammagani-osi:
- https://osi-cwds.atlassian.net/browse/COG-1078

- Added workers phone number in user detail view/edit screens and verify user screens. Also did some refactor to the existing office phone number.


RamuKammagani-osi:
restest this please

RamuKammagani-osi:
retest this please

app/javascript/selectors/detailSelector.js
```diff
@@ -161,9 +161,19 @@ export const assignedRoles = state => {
 }
 
 export const formattedPhoneNumber = state => {
-  const details = selectDetailRecords(state)
-  if (details) {
-    return formatPhoneNumberWithExt(details)
+  const workerPhone = safeGet(state, 'fetchDetails.details.records.user.phone_number', '')
```
Nasir2712: I think we can refactor the logic in both these methods to a separate method and just call it in both the methods.

app/javascript/selectors/validateNewUserSelector.js
```diff
@@ -15,10 +15,20 @@ export const officeName = state => {
   return translateOfficeName(state, officeId)
 }
 
+export const formattedOfficePhoneNumber = state => {
+  const phoneNumber = safeGet(state, 'validateNewUser.verifyUserDetails.user.office_phone_number', '')
+  const ext = safeGet(state, 'validateNewUser.verifyUserDetails.user.office_phone_extension_number', '')
+  if (phoneNumber) {
+    return formatPhoneNumberWithExt(phoneNumber, ext)
```
Nasir2712: same thing here, we can probably just pass in state and the parameters in one method and keep it any of the selector files and we can just export it wherever we want to. This way we can avoid using duplicate logic everywhere.

RamuKammagani-osi: refactored to return object where we can select what we need from object in container.

app/javascript/selectors/detailSelector.js
RamuKammagani-osi: Please see below comment

app/javascript/containers/detailsContainer.js
```diff
@@ -69,7 +69,8 @@ function mapStateToProps(state) {
     assignedRole: assignedRoles(state),
     lastLoginDateTime: lastLogin(state),
     resentRegistrationExistingDateTime: resentRegistrationDate(state),
-    formattedPhoneNumber: formattedPhoneNumber(state),
+    formattedOfficePhoneNumber: formattedPhoneNumber(state).officePhoneNumber,
+    formattedPhoneNumber: formattedPhoneNumber(state).workerPhoneNumber,
```
Sushmareddy-OSI: to maintain naming convention, I would like to change this names to `OfficePhoneNumber` and `WorkerPhoneNumber`. Because , we are not actually formatting at this level, retrieving already formatted value.

app/javascript/containers/addUserContainer.js
```diff
@@ -17,7 +17,8 @@ function mapStateToProps(state) {
     dashboardUrl: '/',
     validateNewUserError: state.validateNewUser.error,
     officeName: officeName(state),
-    officePhoneNumber: formattedPhoneNumber(state),
+    formattedOfficePhoneNumber: formattedPhoneNumber(state).officePhoneNumber,
+    formattedPhoneNumber: formattedPhoneNumber(state).workerPhoneNumber,
   }
```
Sushmareddy-OSI: Same here!

app/javascript/containers/detailsContainer.js

app/javascript/containers/addUserContainer.js
RamuKammagani-osi: Updated!

app/javascript/containers/detailsContainer.js
```diff
@@ -69,7 +69,8 @@ function mapStateToProps(state) {
     assignedRole: assignedRoles(state),
     lastLoginDateTime: lastLogin(state),
     resentRegistrationExistingDateTime: resentRegistrationDate(state),
-    formattedPhoneNumber: formattedPhoneNumber(state),
+    officePhoneNumber: formattedPhoneNumber(state).officePhoneNumber,
+    workerPhoneNumber: formattedPhoneNumber(state).workerPhoneNumber,
   }
```
Nasir2712: I'll leave this comment for you and the rest of them to decide, do you want separate methods for office and worker phone numbers like this below or it's the fine the way it's already is.
```
officePhoneNumber: officePhoneNumber(state)
workerPhoneNumber: workerPhoneNumber(state)
```

RamuKammagani-osi: I feel like Instead of having two selectors, having one selector which returns the expected result will be better unless we have an issue by doing so.

PR #281, opened 2019-01-07
Cog 1066 display change log on details page

Sushmareddy-OSI:
[COG-1066, ChangeLog](https://osi-cwds.atlassian.net/browse/COG-1066)
- Permissions and ChangeLog table in userDetail page


Sushmareddy-OSI:
In DropDown Component {Select} is imported from to react-select. The library is updated to use components for changeLog, due to this update there is also update included for { Select } (react-select v2.1.1). 

According to the react-select-v2.1.1, need to change selectedOption and onChange() methods. This will be addressed in another PR

app/javascript/common/DropDown.js
```diff
@@ -1,6 +1,6 @@
 import React from 'react'
 import PropTypes from 'prop-types'
-import { Select } from '@cwds/components'
+import Select from 'react-select'
```
kelfink: is there a problem with the component library Select ?

app/javascript/common/DropDown.test.js
```diff
@@ -38,9 +38,4 @@ describe('DropDown', () => {
     expect(instance.props.simpleValue).toEqual(true)
     expect(instance.props.disabled).toEqual(false)
   })
-
-  it('returns the Input and Label tags', () => {
```
kelfink: If the input and label are going away, isn't something else replacing them?

app/javascript/common/DropDown.js
Sushmareddy-OSI: In component Library they updated to react-select 2.1.1, according to the update we need to do some changes for `onChange` and `value`(selectedOption). I working on it, will do another PR to update meanwhile I don't wanna block this story.(changeLog)

app/javascript/common/DropDown.test.js
Sushmareddy-OSI: I removed this when tried updating the componentLibrary, that's what I need to figure that out what replaced and there are some prop names changed as well. 
https://react-select.com/upgrade-guide

sunilb0575:
Can you please add two test cases, for testing render of both new cards(Change Log & Permissions from CWS-CMS)?
ex:
 ```it('renders component "PageHeader"', () => {
      expect(wrapper.find('ChangeLog').length).toBe(1)
    })
```

File: app/javascript/views/userDetail/ChangeLog.jsx [hunk 3]
```diff
@@ -0,0 +1,79 @@
...
+  },
+  {
+    Header: 'Notes & Details',
+    accessor: 'notes',
+    minWidth: 50,
+  },
+]
+
+const data = [
+  { time_stamp: 'Thu Mar 03 2018 14:22:43', type: 'Permission Changes', admin_name: 'H' },
```
kelfink: We need to think about this time_stamp value.  It doesn't sort correctly, even though this is fake data.  Is it the format we'll receive it in?

Sushmareddy-OSI: I don't think so and I am not sure in which format will receive the time-stamp. I just represented this way since we are formatting other time related data-fields to show in this format on the list and details page. That may change based on the design in next iteration.

Sushmareddy-OSI:
> Can you please add two test cases, for testing render of both new cards(Change Log & Permissions from CWS-CMS)?
> ex:
> 
> ```
> it('renders component "ChangeLog"', () => {
>      expect(wrapper.find('ChangeLog').length).toBe(1)
>    })
> ```

app/javascript/views/userDetail/CWSPermissions.test.js
```diff
@@ -0,0 +1,17 @@
...
+import { shallow } from 'enzyme'
+import CWSPermissions from './CWSPermissions'
+
+describe('CWSPermissions', () => {
+  let wrapper
+  beforeEach(() => {
+    wrapper = shallow(<CWSPermissions />)
+  })
+
+  it('renders the Components ', () => {
```
kelfink: I like having only one expect per 'it' block.  In this case, it seems like we should really just be testing that there's a rolodex (the dropdown) and further testing that there's the intended card title.  This way the test isn't asking 'how' this card/header/title scheme works, but that the important output is happening.  what do you think about splitting it into two it blocks testing for Rolodex in the first, and  CardTitle with "Permissions from CWS-CMS" text in the second?

app/javascript/views/userDetail/ChangeLog.jsx
kelfink: I think when the actual API works we can assume it's going to have format like " "last_login_date_time": "2019-01-08 14:52:49". and that we should use DataGrid (react-table's) accessor / id style sorting mechanism.  See UserList.jsx and 

```suggestion
  const formatDate = row => {
  return checkDate(row.time_stamp)
}

...

const columnConfig = [
  {
    Header: 'Date/Time',
    id: 'time_stamp',
    accessor: formatDate,
    minWidth: 75,
  },
  
  ...
```

PR #283, opened 2019-01-09
fix changeLog style

Sushmareddy-OSI:
[-COG-1066](https://osi-cwds.atlassian.net/browse/COG-1066)

-Style fix for ChangeLog Component

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -17,7 +17,20 @@
 @import 'bootstrap-sprockets';
 @import 'bootstrap';
 @import 'react-wood-duck/dist/styles/application.scss';
-@import '@cwds/core/dist/styles.css';
+@import '@cwds/core/dist/styles';
+
+body {
+  background-color: #e8e8e8;
+  color: #050505;
+  font-family: ‘Source Sans Pro’, ‘Helvetica Neue’, ‘Helvetica’, ‘Roboto’, ‘Arial’, sans-serif;
+  font-size: 1.5rem;
+  line-height: 1.428571429;
+}
+
+label {
+  display: block;
+  margin-bottom: 0.5rem;
+}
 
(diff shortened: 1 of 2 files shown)


app/javascript/views/usersList/UsersList.scss
```diff
@@ -35,7 +35,7 @@
     .-previous,
     .-next {
       flex: 1 1 40px;
-      margin-top: 12.5px;
+      margin-top: 5.5px;
```
kelfink: is this to fix the little box showing near the dropdown in safari?

app/assets/stylesheets/application.scss
```diff
@@ -17,7 +17,20 @@
 @import 'bootstrap-sprockets';
 @import 'bootstrap';
 @import 'react-wood-duck/dist/styles/application.scss';
-@import '@cwds/core/dist/styles.css';
+@import '@cwds/core/dist/styles';
+
+body {
```
kelfink: this looks very global.  affects all pages, doesn't it ?

Sushmareddy-OSI: It is overriding with the component library styles.

app/javascript/views/usersList/UsersList.scss
Sushmareddy-OSI: No, with the new update of component library, the styles for pagination is breaking. This is for the previous and next buttons in pagination

kelfink: oh, of course "-next"

app/assets/stylesheets/application.scss
kelfink: Ok.    It does seem strange we need to do this.  I wonder if the component library is intent on changing our styles for our apps, or if it should not be really using styles on body.  @eckdanny-osi might weigh in here when available.

PR #285, opened 2019-01-15
Cog 1125 Edit and Save Phone number in user detail screens.

RamuKammagani-osi:
- https://osi-cwds.atlassian.net/browse/COG-1075

- Edit functionality in user detail screen, Able to edit and save phone numbers.


RamuKammagani-osi:
retest this please

app/javascript/_utils/formatters.test.js
```diff
@@ -16,6 +16,13 @@ describe('#formatPhoneNumberWithExt', () => {
       const phoneExtensionNumber = undefined
       expect(formatPhoneNumberWithExt(phoneNumber, phoneExtensionNumber)).toEqual('(111) 444-5555 Ext')
     })
+
+    it('returns phone & Extension with Ext text when ext is 0', () => {
```
Nasir2712: need one more test case for else condition

app/javascript/selectors/detailSelector.test.js
```diff
@@ -709,4 +761,77 @@ describe('selectors', () => {
       })
     })
   })
+
+  describe('#unformattedPhoneNumber', () => {
+    it('returns phone number if phone_number is 11 digits ', () => {
```
Nasir2712: I think you can move this it block to the 
```describe when phone is 10 digits block```
that would make this it block more sense.

File: app/javascript/selectors/detailSelector.test.js [hunk 3]
```diff
@@ -709,4 +761,77 @@ describe('selectors', () => {
...
+      const state = { fetchDetails: undefined }
+      expect(unformattedPhoneNumber(state)).toEqual('')
+    })
+  })
+
+  describe('#isPhoneNumberValid', () => {
+    it('return true when phone number is valid', () => {
+      const state = getState({ phoneNumber: '9163334444' })
+      expect(isPhoneNumberValid(state)).toEqual(true)
+    })
```
Nasir2712: what happens if the phone number is '3344 444'? can we get a small test case for this input as well?

app/javascript/views/userDetail/UserDetailEdit.jsx
```diff
@@ -95,9 +95,28 @@ const UserDetailEdit = ({
...
+              <InputComponent
+                id="InputPhoneNumber"
+                label="Phone Number"
+                fieldClassName="form-group"
+                type="tel"
+                placeholder="Enter numbers only"
+                value={unformattedPhoneNumber}
+                onChange={event => onInputChange('phone_number', event.target.value)}
+                validationError={!isPhoneNumberValid}
+                validationErrorMessage={'Please enter valid phone number'}
```
Nasir2712: I feel like we need to have a constants file name validationErrorMessages so that we could move these strings in there, just because if the message needs tweaking it could just happen in one place.

app/javascript/views/userDetail/UserDetailEdit.jsx
```diff
@@ -95,9 +95,28 @@ const UserDetailEdit = ({
...
+                maxLength={10}
+              />
+            </div>
+            <div className="col-md-1">
+              <InputComponent
+                id="InputPhoneNumberExtension"
+                label="Ext"
+                fieldClassName="form-group"
+                type="number"
+                value={details.phone_extension_number || ''}
```
Nasir2712: can this ``` || ``` logic be moved somewhere in the selector? i feel like there would be an unnecessary condition at the time of rendering.

app/javascript/selectors/detailSelector.js
```diff
@@ -69,7 +69,7 @@ export const isEmailValid = state => {
 }
 
 export const disableActionButton = state => {
-  if (!isEmailValid(state)) {
+  if (!isEmailValid(state) || !isPhoneNumberValid(state)) {
```
Nasir2712: I have a concern with this logic. It will become more complex as we move on when we add more input fields, in order to check the disable button we will keep adding conditional checks. We have to find a way to keep this from happening. I suggest we look at intake's project on how they handle this or come up with a better solution.

app/javascript/selectors/detailSelector.js
```diff
@@ -160,11 +165,25 @@ export const assignedRoles = state => {
   return formatSelectedRoles(assignedRole, rolesList(state))
 }
 
-export const formattedPhoneNumber = state => {
+export const unformattedPhoneNumber = state => {
   const workerPhone = safeGet(state, 'fetchDetails.details.records.user.phone_number', '')
+  if (workerPhone && workerPhone.length === 11) {
+    const phone = workerPhone.substring(1)
```
Nasir2712: any reasons why just remove only the first character in the string?

RamuKammagani-osi: The API is sending Phone number with country code by removing just '+' sign from it, 
if phone number is like Ex: +1 (916) 000-0000 , then api returning as 19160000000, So thats why im removing the first character from it.

app/javascript/_utils/formatters.test.js
RamuKammagani-osi: Sure will update

app/javascript/selectors/detailSelector.test.js
RamuKammagani-osi: Sure

app/javascript/selectors/detailSelector.js
RamuKammagani-osi: Sure

app/javascript/_utils/formatters.test.js
RamuKammagani-osi: Done!

app/javascript/selectors/detailSelector.test.js
RamuKammagani-osi: Done!

app/javascript/selectors/detailSelector.test.js
RamuKammagani-osi: Added!

app/javascript/views/userDetail/UserDetailEdit.jsx
RamuKammagani-osi: Updated!

app/javascript/views/userDetail/UserDetailEdit.jsx
Nasir2712: ok, can you create a separate story for this?

app/javascript/selectors/detailSelector.js
Nasir2712: Also, if you don't mind a tech debt story for this approach as well.

app/javascript/views/userDetail/UserDetailEdit.jsx
```diff
@@ -95,9 +96,28 @@ const UserDetailEdit = ({
...
           <div className="row">
             <div className="col-md-3">
-              <ShowField label="Phone Number">
-                <span>{workerPhoneNumber}</span>
-              </ShowField>
+              <InputComponent
+                id="InputPhoneNumber"
+                label="Phone Number"
+                fieldClassName="form-group"
+                type="tel"
```
Sushmareddy-OSI: Change type to "number", so it allows only numbers. Currently it's allowing to type alpha-characters

RamuKammagani-osi:
retest this please

PR #286, opened 2019-01-16
COG 1110 detail view changelog api

kelfink:
- https://osi-cwds.atlassian.net/browse/COG-1110

- Adds real (Elastic Search-based) API data to the ChangeLog component


kelfink:
retest this please

Nasir2712:
My assumption is we should review this PR after merging @Sushmareddy-OSI PR just because this might get a bunch of merge conflicts which could potentially break some functionality.

Nasir2712:
Also, any specific reason to add minor label rather than patch?

kelfink:
> Also, any specific reason to add minor label rather than patch?

Doesn't a patch imply it's just fixing a bug and doesn't change compatibility?  This change depends on the functionality added to Perry which supplies the changelog data.

PR #289, opened 2019-01-18
Cog 1110 detail view changelog api

kelfink:
- https://osi-cwds.atlassian.net/browse/COG-1110

- Connect the ChangeLog component to backend data in ES.


lib/users/user_repository.rb
```diff
@@ -19,7 +19,8 @@ def get_users_details(id, token)
       Rails.logger.info "404 response from perry with token #{token}" if response.status == 404
       return {} if response.status == 404
 
-      User.new(response.body)
+      audit_events = get_user_audit_events(id, token)
+      User.new(response.body.merge(auditevents: audit_events))
```
Sushmareddy-OSI: Is there any specific reason to add auditevents in get_user_details ?? 
I would suggest to make a separate call. Because , we have to refresh the page after save to see the updated changeLog, right now we are not doing fetchDetails call on save. 

any other thoughts ???

Sushmareddy-OSI: Is there any specific reason to add auditevents in get_user_details ?? 
I would suggest to make a separate call. Because , we have to refresh the page after save to see the updated changeLog, right now we are not doing fetchDetails call on save. 

any other thoughts ???

kelfink:
retest this please

lib/users/user_repository.rb
kelfink: I didn't realize we're not fetching after the modification.  It was much easier to attach the audit logs to the detail, as I was thinking we always look at it at the same time as the details.  It removed the need for a separate saga, etc...  I'm not sure that fetching the audit log separately would fix that aspect.  I think there's a delay between the Save and when the ES index is updated.

kelfink: I've added more commits and changed some behavior based on this comment.  The sage saving details now triggers the retrieval saga so the updated record is brought back on-screen.  This will take care, not only, of the ChangeLog, but the potential for the back-end to update fields like cognito registration status, without us knowing.

kelfink: Code has been updated to perform the getDetails saga after the save, so we have both the updated changeling *and* any follow-on changes that the back-end has performed, such as the user's status changing due to Cognito rules.

app/javascript/views/userDetail/ChangeLog.test.js
```diff
@@ -1,11 +1,30 @@
...
+    event_type: 'C',
+    admin_name: 'wrong1',
+    timestamp: '2019-01-04 14:22:22',
+  })
+  events.push({
+    event: { admin_name: 'Walsh, Jack', admin_role: 'County Admin' },
+    event_type: 'A',
+    admin_name: 'wrong3',
+    timestamp: '2019-01-02 14:22:22',
+  })
```
sunilb0575:
```suggestion
  events = [
  {
    event: { admin_name: 'Dorfler, Marvin', admin_role: 'Office Admin' },
    event_type: 'B',
    admin_name: 'wrong2',
    timestamp: '2019-01-03 14:22:22',
  },
  {
    event: { admin_name: 'Mosely, Alonso', admin_role: 'State Admin' },
    event_type: 'C',
    admin_name: 'wrong1',
    timestamp: '2019-01-04 14:22:22',
  },
  {
    event: { admin_name: 'Walsh, Jack', admin_role: 'County Admin' },
    event_type: 'A',
    admin_name: 'wrong3',
    timestamp: '2019-01-02 14:22:22',
  }
  ]
```

lib/users/user_repository.rb
```diff
@@ -80,8 +81,31 @@ def self.search_base_url
       Rails.configuration.micro_services['base_search_api_url']
     end
 
+    def get_user_audit_events(id, auth_header)
+      http_search_service = Infrastructure::HttpService.new(UserRepository.search_base_url)
+      response = http_search_service.post('/dora/auditevents/auditevent/_search',
```
RamuKammagani-osi: Why can't we use the instance variable @http_service here, like other methods??

Nasir2712:
screen shot 2019-01-22 at 2 15 11 pm
When click on the view I get this undefined on the title of the modal

lib/users/user_repository.rb
kelfink: A:  the @http_service points us to the host for security access-- Perry.  This one needs to goto our Elastic search service, Dora.  I might refactor later so both services are injected the same way (see constructor)

app/javascript/selectors/detailSelector.js
```diff
@@ -16,6 +16,12 @@ export const selectDetailRecords = state => {
   return userDetails
 }
 
+export const selectAuditEvents = state => {
+  const userData = selectUserDetailObject(state)
```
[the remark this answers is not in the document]
kelfink: export ??

PR #294, opened 2019-01-23
Cog 1141 change log details design change

Sushmareddy-OSI:
[https://osi-cwds.atlassian.net/browse/COG-1141](COG-1141)

- ChangeLog Details view new design with actual data


app/javascript/_utils/faIcons.js
```diff
@@ -0,0 +1,10 @@
+import React from 'react'
+import { library } from '@fortawesome/fontawesome-svg-core'
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
+import { faTimesCircle } from '@fortawesome/free-solid-svg-icons'
+
+library.add(faTimesCircle)
+
+export const Circle = () => {
+  return <FontAwesomeIcon icon="times-circle" />
+}
```
sunilb0575: Instead of writing all this code, We can include
```<i className="fa fa-times-circle fa-lg" aria-hidden="true" />``` inside button component.

ratneshraval: @Sushmareddy-OSI  can you verify and correct if needed? Less code is better for all of us.

PR #297, opened 2019-01-24
Publishes port for redis so it can work locally

edgibbs:
Added because when you run redis locally from docker-compose, nothing can access it
from the outside.


edgibbs:
Well the issue is mostly that you can't run a bunch of this locally by default, it was hard to get the spike up and running.  But I'll go ahead and close this one for now.

PR #302, opened 2019-01-29
COG-1169 Remove Card level buttons and add buttons to page header.

RamuKammagani-osi:
- https://osi-cwds.atlassian.net/browse/COG-1169

- Updated design of buttons in page header and if the user is editable, renders the user detail edit page directly. If not editable renders view mode.


app/javascript/views/usersList/UsersList.test.js
```diff
@@ -98,13 +96,13 @@ describe('UsersList', () => {
...
       )
-      expect(wrapperLocal.find('Cards').props().cardHeaderButton).toBe(true)
       expect(wrapperLocal.find('Cards').props().cardHeaderText).toBe('State Administrator view')
-      expect(wrapperLocal.find('Cards').props().headerBtnName).toBe('+ Add a user')
     })
 
     it('renders PageHeader component', () => {
+      //  const buttonValue = wrapper.find('PageHeader').props().button
       expect(wrapper.find('PageHeader').length).toBe(1)
+      // expect(buttonValue).toBe('+ Add a user')
```
Sushmareddy-OSI: please remove these commented lines

RamuKammagani-osi: Done!

app/javascript/common/PageHeaderButtons.test.js
```diff
@@ -0,0 +1,68 @@
...
+      wrapper
+        .find('Button')
+        .at(1)
+        .props().onClick
+    ).not.toThrow()
+  })
+
+  it('renders only one Button when user is not editable', () => {
+    wrapper.setProps({ isUserEditable: false })
+    expect(wrapper.find('Button').length).toBe(1)
```
kelfink: I think this line is obsolete.  If the next line passes, this one is necessarily a length == 1, otherwise props can't be grabbed

app/javascript/reducers/detailsReducers.test.js
```diff
@@ -18,14 +18,15 @@ describe('reducer', () => {
...
+    const state = { details: null, fetching: true, fetchDetailsError: null, resendEmailUserId: ['SOME_ID'] }
     const records = { XHRStatus: 'ready', records: { id: 'key1', username: 'user1' } }
 
     expect(fetchDetails(state, responseAction)).toEqual({
       fetching: false,
       details: records,
-      isEdit: false,
       fetchDetailsError: null,
+      disableActionBtn: true,
+      resendEmailUserId: ['SOME_ID'],
```
kelfink: Just adding coverage here? looks like the implementation didn't change

app/javascript/common/PageHeaderButtons.test.js
```diff
@@ -0,0 +1,68 @@
...
+    expect(wrapper.find('Button').props().disabled).toBe(true)
+    expect(
+      wrapper
+        .find('Button')
+        .at(0)
```
kelfink: see if you can remove this at(0). and maybe use "first" instead of "find" to make the code simpler.  I think there's a way to find the first match rather than all, to simplify the code.

app/javascript/reducers/detailsReducers.test.js
RamuKammagani-osi: Yes, i noticed that was missing and added.

app/javascript/selectors/detailSelector.js
```diff
@@ -67,7 +67,7 @@ export const disableActionButton = state => {
   return safeGet(state, 'fetchDetails.disableActionBtn')
 }
 
-export const checkEditDisable = state => !safeGet(state, 'fetchDetails.details.records.edit_details.editable')
+export const isUserEditable = state => safeGet(state, 'fetchDetails.details.records.edit_details.editable') || false
```
kelfink:
```suggestion
export const isUserEditable = state => safeGet(state, 'fetchDetails.details.records.edit_details.editable', false)
```
I think the semantics are the same, where safeGet gives you false if the expression can't be found.

RamuKammagani-osi: Yeah instead i'm grabbing the value and making use of it! Both works as well

app/javascript/selectors/detailSelector.test.js
```diff
@@ -318,25 +318,25 @@ describe('selectors', () => {
...
+    it('returns false if editable is null', () => {
       const state = getState({ isDetailsEditable: null })
-      expect(checkEditDisable(state)).toEqual(true)
+      expect(isUserEditable(state)).toEqual(false)
     })
 
     it('return true if editable is undefined ', () => {
       const state = getState({ isDetailsEditable: undefined })
-      expect(checkEditDisable(state)).toEqual(true)
+      expect(isUserEditable(state)).toEqual(false)
```
kelfink: Might want a case where isUserEditable doesn't appear in the document.  You have a lot of cases covered though

app/javascript/views/userDetail/UserDetailEdit.jsx
```diff
@@ -2,16 +2,13 @@ import React from 'react'
 import PropTypes from 'prop-types'
 import Cards from '../../common/Card'
 import ShowField from '../../common/ShowField'
-import { InputComponent } from 'react-wood-duck'
+import { InputComponent, Button } from 'react-wood-duck'
```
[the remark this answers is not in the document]
RamuKammagani-osi: There's another coming for editing Notes, I'll try to update in it.

app/javascript/common/PageHeaderButtons.test.js
RamuKammagani-osi: Updated!

PR #306, opened 2019-01-30
new line added for redeploying to int

Sushmareddy-OSI:
-N/A

-rebuild jenkins test


README.md
```diff
@@ -23,6 +25,8 @@ Things you may want to cover:
 
 * Pull Requests and code climate
 
+
 # Questions
 
+
```
NagaPradeepGajula-osi: all of these are unnecessary spacing. Better to add a new line at end of the file.
Also the PR title should be changed we aren't updating anything here. I would suggest the title to be `new line added for redeploying to int`

PR #307, opened 2019-01-31
Adds optional parameter when rebuilding for Jenkins UI master builds

edgibbs:
It was mentioned that couldn't retrigger the build so added a param to handle it.


Jenkinsfile
```diff
@@ -28,6 +28,11 @@ def node_to_run_on() {
 
 node(node_to_run_on()) {
   def app
+  properties([
+    parameters([
+      string(name: 'INCREMENT_VERSION', defaultValue: '', description: 'major, minor, or patch')
+    ])
+  ])
```
kelfink: so this creates the parameter INCREMENT_VERSION with a default of empty string, right?  (groovy weirdness)

Jenkinsfile
```diff
@@ -41,7 +46,7 @@ node(node_to_run_on()) {
       }
 
       stage('Increment Tag') {
-        newTag = newSemVer()
+        newTag = newSemVer(env.APP_VERSION)
```
kelfink: How is this related to the original problem, or the param added above?  Not sure which of these two changes fixes the problem.

PR #308, opened 2019-01-31
Update sorting on changelog, use secondary sorting by admin name + timestamp

kelfink:
- https://osi-cwds.atlassian.net/browse/COG-1112

- fixes the sort order for the change log on user detail


app/javascript/views/userDetail/ChangeLog.jsx
```diff
@@ -3,15 +3,23 @@ import PropTypes from 'prop-types'
 import { Rolodex, Card, CardBody, CardHeader, CardTitle, DataGrid } from '@cwds/components'
 import ModalComponent from './Modal'
 import { checkDate } from '../../_utils/formatters'
+import safeGet from 'lodash.get'
+
+export const sortByName = (a, b) => {
```
RamuKammagani-osi: can you move this to selector, instead of having this in component itself?

app/javascript/views/userDetail/ChangeLog.test.js
```diff
@@ -74,12 +80,70 @@ describe('ChangeLog', () => {
...
+    it('breaks a tie with timestamp', () => {
+      const rowA = {
+        event: { admin_name: 'apple' },
+        timestamp: '2050-01-01 12:00:00',
+      }
+      const rowB = {
+        event: { admin_name: 'apple' },
+        timestamp: '2000-01-01 12:00:00',
+      }
+      expect(rowA.event.admin_name).toEqual('apple')
```
RamuKammagani-osi: can you add specs for undefined and null cases like if admin_name's are null or undefined!

app/javascript/views/userDetail/ChangeLog.jsx
Nasir2712: I think it's fine if they stay here in this component just because no state is being modified here it's just sorting.

PR #310, opened 2019-02-04
Fix for the Jenkins master build

AVandana-OSI:
- No Jira Story

- Added the trigger properties to fix the Jenkins master build and not to override the token and post parameters.


kelfink:
Would https://github.com/ca-cwds/county-admin/blob/master/JenkinsfileAcceptance need similar fixes?  Or Arte they based only on needing downstream triggers to work? @AVandana-OSI

AVandana-OSI:
@kelfink Jenkins acceptance should be fine, trigger propties are for PR and master build.

PR #311, opened 2019-02-04
COG-520 Edit and save notes for a user.

RamuKammagani-osi:
- https://osi-cwds.atlassian.net/browse/COG-520

- Edit and save notes for a user, Displays the no. of characters used while editing a user.


app/assets/stylesheets/application.scss
```diff
@@ -52,6 +52,18 @@ input[type=number]::-webkit-outer-spin-button {
   text-transform: none;
 }
 
+.text-area {
+  font-size: 1.5rem !important;
```
kelfink: are all these !important bits necessary?  Usually implies there's a conflict between this and another style

app/javascript/selectors/detailSelector.js
```diff
@@ -181,3 +182,8 @@ export const formattedPhoneNumber = state => {
 export const phoneExtension = state => {
   return safeGet(state, 'fetchDetails.details.records.user.phone_extension_number') || ''
 }
+
+export const userNotes = state => {
+  const notes = safeGet(state, 'fetchDetails.details.records.user.notes') || ''
```
kelfink: can the || '' be changed to
```suggestion
  const notes = safeGet(state, 'fetchDetails.details.records.user.notes', '')
```
?

app/javascript/views/userDetail/Notes.jsx
```diff
@@ -1,20 +1,38 @@
...
+        {isUserEditable ? (
+          <div>
+            <Input
+              type="textarea"
+              className="form-control text-area"
+              value={userNotes}
+              onChange={event => onChange('notes', event.target.value)}
+              maxLength={250}
+            />
+            <div className="pull-right">Characters used: {`${notesWordCount}/ 250`}</div>
```
kelfink:
```suggestion
            <div className="pull-right">Characters used: {`${notesWordCount}/250`}</div>
```
for consistent spacing.  It's just my liking though.

app/assets/stylesheets/application.scss
Nasir2712: Ya right, the funny thing is having wood duck styles and component library styles are different. They create conflicts and that is why we have to forcibly apply our own styles

app/javascript/views/userDetail/UserDetail.test.js
```diff
@@ -208,10 +208,10 @@ describe('UserDetail', () => {
 
     describe('renders cards', () => {
       it('should display <UserDetailShow/>', () => {
-        wrapper.setProps({ isUserEditable: false })
-        expect(wrapper.find('UserDetailShow').length).toBe(1)
         wrapper.setProps({ isUserEditable: true })
         expect(wrapper.find('UserDetailShow').length).toBe(0)
+        wrapper.setProps({ isUserEditable: false })
```
kelfink: Not sure why the order of these was changed, but they should independent.  Would be better to stitch to two separate 'it' blocks....
```suggestion
        it('should display <UserDetailShow/> when user is not editable', () => {
	        wrapper.setProps({ isUserEditable: false })
	        expect(wrapper.find('UserDetailShow').length).toBe(1)
	      });
	              it('should not display <UserDetailShow/> when user is editable', () => {
	        wrapper.setProps({ isUserEditable: true })
	        expect(wrapper.find('UserDetailShow').length).toBe(0)
	      });
```

app/javascript/views/userDetail/UserDetail.test.js
```diff
@@ -220,16 +220,20 @@ describe('UserDetail', () => {
       })
 
       it('should display <UserDetailEdit/>', () => {
-        wrapper.setProps({ isUserEditable: true })
-        expect(wrapper.find('UserDetailEdit').length).toBe(1)
```
kelfink: split into two it blocks as suggested above in the other case.

app/assets/stylesheets/application.scss
RamuKammagani-osi: @kelfink As @Nasir2712 said we have to force the styles to match with our current styles.

app/javascript/selectors/detailSelector.js
RamuKammagani-osi: @kelfink  That works when notes are undefined, But when notes are null it still gives me the null while testing the selector, So that's why i used that || '' to return '' if notes are null.

app/javascript/views/userDetail/UserDetail.test.js
RamuKammagani-osi: Updated!

app/javascript/views/userDetail/UserDetail.test.js
RamuKammagani-osi: Updated!

RamuKammagani-osi:
@kelfink Updated the changes you requested in the b0c3a64

app/assets/stylesheets/application.scss
kelfink: Ok.  It's just a code smell.  stuck with it for now until we ditch wood duck completely

app/javascript/selectors/detailSelector.js
kelfink: cool, good enough of a reason.

PR #313, opened 2019-02-04
Cog 1203 import buttons from component library

Sushmareddy-OSI:
[https://osi-cwds.atlassian.net/browse/COG-1203](https://osi-cwds.atlassian.net/browse/COG-1203)

- replace react-wood-duck buttons with @cwds/components in 

- [x] Verify and add User card buttons
- [x]  ChangeLog view button
- [x] resend Invite button


app/assets/stylesheets/application.scss
```diff
@@ -38,8 +38,25 @@ input[type=number]::-webkit-outer-spin-button {
   margin: 0;
 }
 
-.col-xs-5 {
-  width: 40%;
+.card-buttons,
+.resend-email-btn {
```
Nasir2712: same over-riding issue with wood-duck and component library as well? that's why had to add CSS for this?

Sushmareddy-OSI: yes, but I am expecting this styling part at-least minimum styles like colors, font-size , classNames and font-weights etc... should be in our @cwds/components. Until then, this is how we need to do (overriding).

We can't get rid of react-wood-duck completely right now because there are other styles and components like Input, card, global header, page header which are most important are from react-wood-duck.

PR #319, opened 2019-02-08
Code 146 resolve cc issues - part 1

NagaPradeepGajula-osi:
-  https://osi-cwds.atlassian.net/browse/CODE-148

- Resolve code climate code smells and duplication for County Admin.


app/javascript/_utils/checkAdminRoles.js
```diff
@@ -2,9 +2,7 @@ export const getAdminOfficeIDs = account => {
   if (
     account &&
     account.admin_office_ids &&
-    !isRoleAvailable(account, 'Super-admin') &&
-    !isRoleAvailable(account, 'State-admin') &&
-    !isRoleAvailable(account, 'County-admin') &&
+    !isRoleAvailable(account, 'Super-admin', 'State-admin', 'County-admin') &&
```
kelfink: not equivalent.  The final two parameters are going to be ignored (thanks, javascript for not checking params)

app/javascript/views/userDetail/UserDetail.jsx
```diff
@@ -49,42 +48,31 @@ export default class UserDetail extends Component {
...
+        <div>{userDetailError ? <UserMessage errorMsg={userDetailError} /> : <UserMessage successMsg={saveSuccessMsg} />}</div>
+      )
     return null
   }
 
-  showAddAlert = () => {
-    if (this.props.id) {
-      return (
+  showAddAlert = () => (
+    <div>
```
kelfink: why not return this?  Don't we need to?

app/javascript/views/userDetail/UserDetail.jsx
```diff
@@ -49,42 +48,31 @@ export default class UserDetail extends Component {
...
+  )
 
-  emailSent = () => {
-    if (this.state.resendEmailAlert) {
-      return <UserMessage successMsg={'Registration email has been sent successfully'} />
-    }
-    return null
-  }
+  emailSent = () => (
+    <div>{this.state.resendEmailAlert ? <UserMessage successMsg={'Registration email has been sent successfully'} /> : null}</div>
```
kelfink: no return ?

NagaPradeepGajula-osi: in this case ternary defaultly returns based on the condition. so wouldn't need the return. verified with tests and on UI. the usermessage is being returned.

app/javascript/views/userDetail/UserDetail.jsx
NagaPradeepGajula-osi: that is what we defaultly return 
screen shot 2019-02-13 at 11 24 02 am

app/javascript/_utils/checkAdminRoles.js
NagaPradeepGajula-osi: nice catch. verified with tests and UI was working fine. Didn't thought of the basics will refactor.

kelfink: If yarn:test was still working we should probably enhance that.

app/javascript/views/userDetail/UserDetail.jsx
kelfink: Ok.  JS doesn't by default return the last expression evaluated (in my knowledge). so some magic is happening there.

PR #321, opened 2019-02-12
Add endpoint to query ES for audit_events by source and office_ids

kelfink:
- Adds an endpoint /api/audit_events
 - takes one parameter 'q' which is the same query used in user_list api endpoint.  Only processes 'office_ids' query parameter at this point.  

 - takes one parameter 'q' which is the same query used in user_list api endpoint.  Only processes 'office_ids' query parameter at this point.  
 - Is part of the overall display of a collection of the changelog displayed on the user list.


spec/lib/elastic/audit_event_query_builder_spec.rb
```diff
@@ -0,0 +1,136 @@
...
+            must: [
+              { terms: { 'event.office_id.keyword': %w[north south east west] } },
+              { match_phrase_prefix: { 'event_source': 'CAP' } }
+            ]
+          }
+        }
+      }
+      input_query = { "query": [{ "field": 'office_ids', "value": %w[north south east west] }],
+                      "sort": [], "size": 50, "from": 0 }
+      # input_query[:last_name] = 'Smith'
```
RamuKammagani-osi: remove comments

PR #322, opened 2019-02-12
COG-804 Privileges from CWS-CMS

sunilb0575:
-https://osi-cwds.atlassian.net/browse/COG-804

- Permissions from CWS-CMS on user detail page.


sunilb0575:
Jenkins retest this please.

app/javascript/common/TableHeaders.js
```diff
@@ -0,0 +1,13 @@
+export const cwsPermissionsColumnConfig = () => [
+  {
+    Header: 'Category',
+    accessor: 'category',
```
Sushmareddy-OSI: I would like to have this in CWSpermissions.jsx,  It will be ideal to add re-usable components only to the common folder.

if your concern is about the number of lines, then I would suggest to separate but still it should be in details.jsx as CWSpermissionsTableHeader.jsx or something else

sunilb0575: Yeah, my only concern is not to make our component longer, when we have the capability to move such similar constants (all the 3 table headers in ```userList```, ```changeLog``` and now ```cwsPermissions```) to one place.
``Much Cleaner Code```

app/javascript/selectors/detailSelector.test.js
```diff
@@ -174,6 +177,20 @@ describe('selectors', () => {
...
+
+    it('selects the CWS permissions', () => {
+      const state = getState({ CWSPrivileges: CWSPrivileges })
+      expect(selectCWSPrivileges(state)).toEqual(CWSPrivileges)
+    })
+
+    it('selects an empty array when there are no CWS permissions', () => {
+      const state = getState({})
+      expect(selectCWSPrivileges(state)).toEqual([])
+    })
```
Sushmareddy-OSI: can you also add two more scenarios when `CWSPrivileges: undefined` and `CWSPrivileges: null`

kelfink:
Why tag MAJOR?  I think that's for releases.  Minor implies not having backward compatibility due to a new feature.

app/javascript/common/TableHeaders.js
kelfink: But does this get us toward reusability?  Not all tables have the same headers...

kelfink: Also, I don't know if the term privileges is right here.  I think we're talking about permissions.  Synonyms, but let's not use both terms.

app/javascript/selectors/detailSelector.test.js
```diff
@@ -174,6 +177,20 @@ describe('selectors', () => {
...
 
+  describe('#cwsPrivileges', () => {
+    const CWSPrivileges = [{ category: 'Access Authority', privilege: 'Adoptions' }]
+
+    it('selects the CWS permissions', () => {
+      const state = getState({ CWSPrivileges: CWSPrivileges })
+      expect(selectCWSPrivileges(state)).toEqual(CWSPrivileges)
+    })
+
+    it('selects an empty array when there are no CWS permissions', () => {
```
```suggestion
    it('selects an empty array when there are no CWS privileges', () => {
```

app/javascript/common/TableHeaders.js
sunilb0575: Sure I can move table header back to CWsPrivileges. 
In CWS-CMS this are shown as privileges, so after confirming my PO changed it to privileges.

app/javascript/selectors/detailSelector.test.js
sunilb0575: For some reason, safeGet do not check for null, good thing you caught it.

PR #324, opened 2019-02-13
Cog 776 warn admin when permissions empty

Sushmareddy-OSI:
[https://osi-cwds.atlassian.net/browse/COG-776](https://osi-cwds.atlassian.net/browse/COG-776)

- Warning with modal before leaving edit user screen to notify admin no permissions have been added
- renamed `modal => changeLogDetailsView` & 
    `pageHeaderButtons => DetailsPageHeaderButtons`


Sushmareddy-OSI:
retest this please

app/javascript/selectors/detailSelector.js
```diff
@@ -117,6 +117,11 @@ const checkForModifiedDetails = (initialDetails, updatedDetails) => {
   return initialDetails === updatedDetails ? undefined : updatedDetails
 }
 
+export const isPermissionsEmpty = state => {
+  const modifiedDetails = selectDetailRecords(state)
```
kelfink: naming this variable.... do we care / is it necessarily "modified" details?  Or is it just details ?

app/javascript/views/userDetail/NoPermissionsWarning.jsx
```diff
@@ -0,0 +1,64 @@
+import React, { Component } from 'react'
+import PropTypes from 'prop-types'
+import { Modal, ModalBody, ModalFooter, ModalHeader, Button } from '@cwds/components'
+
+export default class ModalComponent extends Component {
```
kelfink: I think this class should be named after the file it's in.  It's not important that it's a ModalComponent (and we might be defining the same name other places, if it were). What we'll want to know down the road is that it's this NoPermissionsWarning object
```suggestion
export default class NoPermissionsWarning extends Component {
```

app/assets/stylesheets/application.scss
```diff
@@ -122,6 +122,15 @@ input[type=number]::-webkit-outer-spin-button {
 
 /* Modal.jsx styles */
```
kelfink: are these common for multiple modal's ?

app/javascript/selectors/detailSelector.js
Sushmareddy-OSI: it just details, updated!!

app/assets/stylesheets/application.scss
Sushmareddy-OSI: No, it depends on design. The modal-body text font-size is different for changeLog Modal and NoPermissionsModal. So we have to add more class-names as per design requirement. Added that comment to know that these are related to **Modal**

kelfink: so I'm wondering why the class is .modal-body-text instead of specific, like nopermissionmodal-body-text

Sushmareddy-OSI: oh!! I didn't thought about the specific class-names.

Sushmareddy-OSI: updated!

kelfink: Yes!  much better thanks!

PR #334, opened 2019-02-20
COG-1067 Add Made To Column in change log

RamuKammagani-osi:
- https://osi-cwds.atlassian.net/browse/COG-1067

- Adds  an extra column if user is on the user list page, also increase the no. of records retrieved for change log in user list.


app/javascript/views/userDetail/ChangeLog.jsx
```diff
@@ -63,7 +70,17 @@ const columnConfig = (
     Cell: row => {
       return `${row.original.event.admin_name} (${row.original.event.admin_role})`
     },
-    sortMethod: sortByName,
+    // sortMethod: sortByName,
```
kelfink: remove comment line

PR #336, opened 2019-02-22
Fix the pagination styles in IE, Edge, fire-fox browsers

Sushmareddy-OSI:
[https://osi-cwds.atlassian.net/browse/COG-1150](https://osi-cwds.atlassian.net/browse/COG-1150)

- Pagination is not aligned in fire-fox, IE , and edge browsers


kelfink:
Since this is refactor and the tests still pass I think we're good

PR #338, opened 2019-02-27
Cog 1244 immediate changeLog visibility

Sushmareddy-OSI:
[https://osi-cwds.atlassian.net/browse/COG-1244](https://osi-cwds.atlassian.net/browse/COG-1244)

- ChangeLog updates on Resend Invite event and Save event


app/javascript/sagas/resendRegistrationEmailSaga.js
```diff
@@ -15,8 +15,10 @@ export function* resendEmail(action) {
       resendEmailStatus,
       id,
     })
-    yield put(fetchDetailsActions(id))
-    yield put(clearAddedUserDetailActions())
+    // reload the audit log for the user
+    yield put(fetchUserAuditEventsActions(id))
+    yield put(clearAuditEvents())
```
kelfink: does this seem like it should come before the fetch?  Maybe not since they operate on different areas of the state.... ?

app/javascript/sagas/saveUserDetailsSaga.js
```diff
@@ -23,7 +25,10 @@ export function* saveDetails(action) {
       saveUserDetails,
       successAlert,
     })
+    // send a message to start a fetch saga.
+    yield put(clearAuditEvents())
     yield put(fetchDetailsActions(saveUserDetailsActions.id))
+    yield put(fetchUserAuditEventsActions(saveUserDetailsActions.id))
```
kelfink: in this sage the clear happens before the new fetch.  not consistent with the previous saga?

app/javascript/selectors/auditEventsSelector.test.js
```diff
@@ -0,0 +1,45 @@
...
+        },
+      },
+    }
+    expect(selectAuditEvents(state)).toEqual({
+      auditEvents: [{ event_source: 'CAP', event_type: 'User Created', timestamp: '2019-01-14 15:10:41' }],
+      county_name: 'County1',
+    })
+  })
+
+  it('returns an empty array when there are no audit events', () => {
```
kelfink: We repeat this pattern a lot.  I wonder if there's a more succinct way to test that all these cases result in the empty array

kelfink: Ok looks like when we go to jest version 23, we can use the it.each method to test several scenarios in a loop....
```  it.each([{}, null, undefined, { auditEvents: null }], 'returns an empty array in empty conditions', val => {
    const state = { fetchAuditEvents: val }
    expect(selectAuditEvents(state)).toEausl([])
  })```

something like that. But not until we upgrade jest.

File: app/javascript/sagas/saveUserDetailsSaga.js (hunk 2 above)
Comment (Sushmareddy-OSI) [reply app/javascript/sagas/saveUserDetailsSaga.js]: this is working when clearing previous auditEvents before we fetch the new auditEvents.

File: spec/acceptance/add_user_spec.rb [hunk 4]
```diff
@@ -107,6 +106,7 @@
 
     # correct the email to a proper address
     email_address = new_email_address
+    puts "Email address for add-user invalid-data testing #{email_address}"
```
kelfink: this output is for debugging purposes down the road.  It's nice to know what email addresses were created by this test run, so we can troubleshoot situations like when the user accounts aren't cleaned up.

PR #343, opened 2019-03-04
COG-1258 Unlock locked users

sunilb0575:
- https://osi-cwds.atlassian.net/browse/COG-1164
- https://osi-cwds.atlassian.net/browse/COG-1258  (FE - Sub-task)

- As an admin unlock locked user, when have edit privileges.


app/javascript/selectors/detailSelector.test.js
```diff
@@ -341,32 +346,131 @@ describe('selectors', () => {
...
     })
 
-    it('return false if editable is not available ', () => {
+    it('returns false when editable & locked are not available ', () => {
       const state = getState({ fetchDetails: {} })
       expect(isUserEditable(state)).toEqual(false)
     })
   })
 
+  describe('#isAdminWithEditPrivileges', () => {
```
RamuKammagani-osi: Can you add more specs here for null and undefined cases, like if ```isDetailsEditable: null & undefined```, we found multiple times that page has broken as we dint cover those two cases in selectors, so just to make sure!

app/javascript/selectors/detailSelector.test.js
```diff
@@ -341,32 +346,131 @@ describe('selectors', () => {
...
+    })
+
+    it('returns undefined if editable is not available', () => {
+      const state = getState({ fetchDetails: {} })
+      expect(isAdminWithEditPrivileges(state)).toEqual(undefined)
+    })
+  })
+
+  describe('#isUserLocked', () => {
+    it('returns true when locked is true', () => {
```
RamuKammagani-osi: same here please

app/javascript/selectors/detailSelector.test.js
sunilb0575: @RamuKammagani-osi as both ```isAdminWithEditPrivileges``` and ```isUserLocked``` return values if any or else returns undefined and as those respective values are used only in ```isUserEditable``` method, I have tested null and undefined in ```isUserEditable``` method. 
And also UI is independent of the return values from those methods(isAdminWithEditPrivileges, isUserLocked).

app/javascript/selectors/detailSelector.test.js
sunilb0575: https://github.com/ca-cwds/county-admin/pull/343#discussion_r262306614

app/javascript/selectors/detailSelector.test.js
RamuKammagani-osi: Alright!, that sounds good.

PR #348, opened 2019-03-11
Cog 1244 immediate changelog visibility

Sushmareddy-OSI:
-https://osi-cwds.atlassian.net/browse/COG-1244

- Immediate ChangeLog updates on changing user-details and resend registration.


app/javascript/sagas/getAuditEventsSaga.js
```diff
@@ -1,10 +1,12 @@
 import { takeLatest, call, put } from 'redux-saga/effects'
+import { delay } from 'redux-saga'
 import * as actionTypes from '../actions/actionTypes'
 import UserService from '../_services/users'
 
 // worker saga: makes the api call when watcher saga sees the action
 export function* getAuditEvents(action) {
   try {
+    yield call(delay, 800)
```
kelfink: Does waiting this one time work out reliably?  I think we may need a way to loop or to always be refreshing this in the background.  
If 800 works very reliably, then we should at least make a comment before this line, since it's a magic number, and no one likes seeing 'sleep' 'wait' or 'delay' type code without an explanation.

app/javascript/sagas/getAuditEventsSaga.test.js
```diff
@@ -35,6 +37,7 @@ describe('sagas', () => {
         const params = {}
         const action = { payload: params }
         const gen = getAuditEvents(action)
+        expect(gen.next().value).toEqual(call(delay, 800))
```
kelfink: Adding this to the test script for now is OK but we might want to refactor the unit tests to test for just what we really want to make sure of.  Let's consider, down the road, the advice here:  https://redux-saga.js.org/docs/advanced/Testing.html. see comments under "Testing the full Saga" as it might make test modifications like this unnecessary.

app/javascript/sagas/getAuditEventsSaga.js
```diff
@@ -1,16 +1,25 @@
...
     yield put({
       type: actionTypes.FETCH_AUDIT_EVENTS_API_CALL_SUCCESS,
       payload: response,
     })
+    // wait for another 3 seconds and fetch again to confirm the updated response
+    yield call(delay, 3000)
+    yield put({
+      type: actionTypes.FETCH_AUDIT_EVENTS_API_CALL_SUCCESS,
+      payload: response,
+    })
```
kelfink: Yeah.  this should do quite well to catch long-delayed entries.  If there are further problems we can do it in a new story, probably designed to continuously poll the API in the background.

PR #350, opened 2019-03-14
COG-1220 Disable hyperlinks for Phone Number 

RamuKammagani-osi:
-  https://osi-cwds.atlassian.net/browse/COG-1220

- Disables hyperlinks for the phone number in add user/ user detail page.


app/javascript/views/userDetail/UserDetail.test.js
```diff
@@ -87,6 +87,12 @@ describe('UserDetail', () => {
     })
   })
 
+  describe('meta tag to disable hyperlinks for the phone number', () => {
+    it('verifies the meta tag exists', () => {
+      expect(wrapper.find('meta').length).toBe(1)
```
kelfink: doesn't seem very useful unless is also finds the specific meta which disables the link.   Can you add the. specific attribute desired, to the find ?

RamuKammagani-osi: Updated!

app/javascript/views/addUser/AddUser.test.js
```diff
@@ -34,6 +34,14 @@ describe('VerifyUser', () => {
     })
   })
 
+  describe('meta tag to disable hyperlinks for the phone number', () => {
+    it('verifies the meta tag exists', () => {
+      expect(wrapper.find('meta').length).toBe(1)
+      expect(wrapper.find('meta').props().name).toBe('format-detection')
+      expect(wrapper.find('meta').props().content).toBe('telephone=no')
```
kelfink: Ideally we just want one test (one expect) per 'it' block.  I think we can cover all of what we want to test here with a more specific find expression...
```suggestion
      expect(wrapper.find("meta[name='format-detection']").props().content).toBe('telephone=no')
```

RamuKammagani-osi: Updated!

PR #352, opened 2019-03-15
Cog 1194 parameterize racfid in acceptance

kelfink:
- https://osi-cwds.atlassian.net/browse/COG-1194

- Instead of hardcoding RACFID in the acceptance tests, pass them in on the spec command, so Jenkins can supply the values differently for each environment.


kelfink:
retest this please

JenkinsDeploy
```diff
@@ -88,13 +88,13 @@ def integrationAcceptenceTest(environment, version){
     stage('Integration Acceptance Test Chrome') {
       sh "docker-compose up -d --build county-admin-test"
       sleep 60
-      sh "docker-compose exec -T --env TZ=${TZ} --env COGNITO_USERNAME=${COG_USER} --env COGNITO_PASSWORD=${COG_PASS} --env COUNTY_AUTHORIZATION_ENABLED=true --env COUNTY_ADMIN_WEB_BASE_URL=https://web.integration.cwds.io/cap county-admin-test bundle exec rspec spec/acceptance --format documentation"
+      sh "docker-compose exec -T --env TZ=${TZ} --env COGNITO_USERNAME=${COG_USER} --env COGNITO_PASSWORD=${COG_PASS} --env ADD_RACFID=${ADD_RACFID}  --env WRONG_COUNTY_RACFID=${WRONG_COUNTY_RACFID} --env COUNTY_AUTHORIZATION_ENABLED=true --env COUNTY_ADMIN_WEB_BASE_URL=https://web.integration.cwds.io/cap county-admin-test bundle exec rspec spec/acceptance --format documentation"
```
RamuKammagani-osi: Why ADD_RACFID??? Why not VALID_RACFID??

spec/acceptance_helper.rb
```diff
@@ -13,7 +13,7 @@
   if ENV['FIREFOX']
     caps = Selenium::WebDriver::Remote::Capabilities.new(accept_insecure_certs: true)
     browser_options = Selenium::WebDriver::Firefox::Options.new
```
RamuKammagani-osi: Line 15 here is not required if we are not running in headless browser!

JenkinsDeploy
kelfink: I think it's more explicit to say it's the RACFID I will "ADD", but also, its in the "contract" for QA.  https://osi-cwds.atlassian.net/browse/COG-1194 see acceptance criteria.  Also, they've made a request of Ops to add these parameters to the Jenkins environment, and I don't want to delay that.   Later we might have other valid RACFID's and they would have different names, also valid.

spec/acceptance_helper.rb
kelfink: Ok I'll remove that.

JenkinsDeploy
RamuKammagani-osi: Sounds fair!

PR #355, opened 2019-03-15
COG-1315-Account locked changelog status

sunilb0575:
- https://osi-cwds.atlassian.net/browse/COG-1297

- Changelog status for locked users


app/javascript/_utils/formatters.test.js
```diff
@@ -187,3 +195,29 @@ describe('#formatPermissions', () => {
...
+    expect(formatMadeByEvent(adminName, adminRole)).toEqual('Some Admin')
+  })
+
+  it('returns only adminName when adminRole is undefined', () => {
+    const adminName = 'Some Admin'
+    const adminRole = undefined
+    expect(formatMadeByEvent(adminName, adminRole)).toEqual('Some Admin')
+  })
+
+  it('returns only adminName when adminRole is null', () => {
```
Nasir2712: what if adminName and adminRole are undefined?

sunilb0575: It will return undefined.

kelfink: Would be nice to issue a test for that.  Also better if it wasn't undefined.  Maybe empty string.  Something we can display without thinking too much about it.

sunilb0575: @kelfink Right now, we are not checking any of the events for null or undefined. I think a new user story should be created in order to determine what to be shown on UI when null or undefined.

kelfink: Sounds like good technical debt, if you write it up.

sunilb0575: Sure.

PR #358, opened 2019-03-20
Use meta tag in page header instead of specific tag for field

RamuKammagani-osi:
- n/a

- Uses meta tag in page headers to disable the hyperlink for the phone numbers in show/edit/add user pages.


app/javascript/views/userDetail/UserDetailShow.test.js
```diff
@@ -109,11 +109,5 @@ describe('UserDetailShow', () => {
           .props().children
       ).toEqual(expectedValue)
     })
-
-    describe('meta tag to disable hyperlinks for the phone number', () => {
```
kelfink: Shouldn't we still have a unit test here checking that the format-detection meta tag is present?

RamuKammagani-osi: its not present in that component right! So it was used in rails layer, i still dint get what you are asking for?

kelfink: Just that someplace we should test that the meta tag shows up.  But I can see where testing it in a component not resolved by a the rails layer won't work,  Ok, maybe I can find a way to test it in a rails unit test.

PR #361, opened 2019-03-25
COG-1320 Change success alert message text when email is edited

RamuKammagani-osi:
- https://osi-cwds.atlassian.net/browse/COG-1320

- Changed the success alert message when email is edited and when not edited it remains same.


app/javascript/common/UserMessage.js
```diff
@@ -3,11 +3,14 @@ import React from 'react'
 import { Alert } from '@cwds/components'
 
 const UserMessage = ({ errorMsg, successMsg }) => {
+  const successWithEmailChange = Array.isArray(successMsg) ? successMsg[1] : ''
+  const successAlertMessage = !Array.isArray(successMsg) ? successMsg : successMsg[0]
```
kelfink:
```suggestion
  const successAlertMessage = Array.isArray(successMsg) ? successMsg[0] : successMsg
```

I think that's equivalent, and much easier to read

app/javascript/common/UserMessage.test.js
```diff
@@ -28,4 +28,10 @@ describe('User Message', () => {
     expect(errorComp.find('Alert').length).toEqual(1)
     expect(errorComp.contains(props)).toBe(true)
   })
+
+  it('verify success block with successMsg when email is edited', () => {
+    const props = ['Successfull', 'Please notify to someone']
+    const errorComp = shallow(<UserMessage successMsg={props} />)
+    expect(errorComp.contains('Successfull', <b>Please notify to someone</b>)).toBe(true)
```
kelfink:
```suggestion
    expect(errorComp.contains('Successful', <b>Please notify to someone</b>)).toBe(true)
```
one l in successful

app/javascript/views/userDetail/UserDetail.jsx
```diff
@@ -239,7 +239,7 @@ UserDetail.propTypes = {
   workerPhoneNumber: PropTypes.string,
   auditEvents: PropTypes.array,
   isPhoneNumberValid: PropTypes.bool,
-  saveSuccessMsg: PropTypes.string,
+  saveSuccessMsg: PropTypes.oneOfType([PropTypes.string, PropTypes.array]),
```
kelfink: Can we indicate here that it's an array if, in the component, we test for times when it is not ?

RamuKammagani-osi: yes we should if not there is a console warning ```Failed PropType:  array was passed instead of string```, Also i have updated the test cases for both string and array.

app/javascript/common/UserMessage.test.js
RamuKammagani-osi: Typo mistake! Updated it just now

app/javascript/common/UserMessage.js
RamuKammagani-osi: Updated just now

PR #363, opened 2019-03-27
COG-1262 Change log should not update when navigating through pages in user list

RamuKammagani-osi:
-  https://osi-cwds.atlassian.net/browse/COG-1262

- Change will not be updated when user is navigating through pages, Change log will be updated when office is selected and clicked on search button.


app/javascript/sagas/searchUsersSaga.js
```diff
@@ -1,14 +1,12 @@
...
 import { getSearchParams } from '../selectors/userListSelector'
 import { searchUsers } from '../actions/userListActions'
-import { fetchAuditEventsActions } from '../actions/auditEventActions'
 import * as actionTypes from '../actions/actionTypes'
 
 export function* doSomething(action) {
   try {
     const params = yield select(getSearchParams)
     yield put(searchUsers(params))
-    yield put(fetchAuditEventsActions(params))
```
kelfink: With this removed, we won't get searched by office to work anymore, will we?  I think we need it.

RamuKammagani-osi: No! We don't need that anymore in searchUsersSaga, because searchUsersSaga is called everytime when user navigates through pages and search for last name. So instead of that i'm calling  fetchAuditEventsActions in handle search, when user selects the office name and search for users then auditEventSaga triggers and updates but not when navigating through pages.

app/javascript/sagas/searchUsersSaga.js
```diff
@@ -1,14 +1,12 @@
 import { takeLatest, select, put } from 'redux-saga/effects'
 import { getSearchParams } from '../selectors/userListSelector'
 import { searchUsers } from '../actions/userListActions'
-import { fetchAuditEventsActions } from '../actions/auditEventActions'
 import * as actionTypes from '../actions/actionTypes'
 
 export function* doSomething(action) {
```
Nasir2712: can we change the name of this saga?

RamuKammagani-osi: Sure! Updated it just now.

PR #372, opened 2019-04-09
User ES queries allow first name and last name with 'should' matching

kelfink:
- https://osi-cwds.atlassian.net/browse/COG-1339

- Allow search by first name in addition to last name.  Make searching a combination (union of first and last matches) sorted by relevancy.


kelfink:
This PR will not complete the story, but it allows other stories to be worked on which also enhance the User ES query.

kelfink:
I pushed up some changes to sort by score, as the story requests.  (closest matches first)

PR #375, opened 2019-04-10
Cog 1393 update component library

Sushmareddy-OSI:
[COG-1393](https://osi-cwds.atlassian.net/browse/COG-1393)

-Updated to latest component library, button component changed as primitiveButton and checkbox component is added within the application since it is deprecated in the latest version of component library.


app/javascript/views/userDetail/ChangeLogDetailsView.jsx
```diff
@@ -1,6 +1,6 @@
 import React, { Component } from 'react'
 import PropTypes from 'prop-types'
-import { Modal, ModalBody, ModalHeader, Avatar, Button } from '@cwds/components'
+import { Modal, ModalBody, ModalHeader, Avatar, PrimitiveButton as Button } from '@cwds/components'
```
kelfink: couldn't we just keep the PrimitiveButton name and rename the references in this file?  Seems like that is less confusing, for those who have/will work in both libraries.  Just an opinion about that, though.

package.json
```diff
@@ -26,7 +26,8 @@
     "seed-users": "scripts/seed-users"
   },
   "dependencies": {
-    "@cwds/components": "0.6.0",
+    "@cwds/components": "1.2.1",
+    "@cwds/forms": "1.2.1",
```
kelfink: I think we no longer need the forms package, and probably shouldn't include it, since it's not ready for production

Sushmareddy-OSI:
retest this please

app/javascript/views/userDetail/ChangeLogDetailsView.jsx
Sushmareddy-OSI: If we reference as PrimitiveButton in each file, that will be tough to change in every place when it changes again.

kelfink: You are so pessimistic!  And probably right.  OK

PR #376, opened 2019-04-12
COG-1388 create tiles on search page

sunilb0575:
-Subtask to show tiles on the search page(landing page of manage users).
https://osi-cwds.atlassian.net/browse/COG-1372

- show tiles on the search page(landing page of manage users).
- click events will be added in the next subtask.


sunilb0575:
retest this please.

app/javascript/actions/searchUserListActions.js
```diff
@@ -39,6 +39,17 @@ export const setSearch = query => ({
   payload: query,
 })
 
+/**
+ * Set search for tiles
+ * @param {Object[]} query
+ * @param {string} query[].field Identifier for the field on which to search
+ * @param {string|number|boolean} query[].value Value on which to search
```
kelfink: how do these params correlate to the method?  I don't follow it

app/javascript/reducers/searchUserListReducers.js
```diff
@@ -56,6 +62,66 @@ function searchUserListReducer(state = initialValue, { type, payload, error, met
         users: null,
       }
 
+    case actionTypes.GET_ACTIVE_USERS_REQUEST:
+      return { ...state, fetching: true, error: null, query: payload.query }
+
+    case actionTypes.GET_ACTIVE_USERS_SUCCESS:
+      state.searchPageTiles[0].count = payload.meta.total
```
kelfink: Named indices would be safer than magic numbers.  What if we decide to add a tile in the first position and move the others down?  We'd have to renumber everything and chase the values around.

kelfink: This sets the count of users based on the ES response's meta's total.  Does this mean all the 100's of users are retrieved, passed back to the front end, then discarded so we only use the count ?  That seems inefficient, if so.

sunilb0575:
@kelfink Seems like this is the best way to update a value inside an array of objects.

app/javascript/actions/searchUserListActions.js
sunilb0575: Here the params are type and query.

app/javascript/reducers/searchUserListReducers.js
sunilb0575: Agree, changed to named indices.
To make it efficient, passing the resulting users list to the filter list page.

app/javascript/actions/searchUserListActions.js
kelfink: Ok, so it's describing to me that query needs to be of the form {field: "identifier", value: string|number|boolean}

PR #378, opened 2019-04-14
Add first name/last name/ email/ CWSLogin to react layer user queries

kelfink:
- https://osi-cwds.atlassian.net/browse/COG-1393

- Add first name to user search queries


kelfink:
Now includes the email and RACFID fields.  #weekend #couchtime

app/javascript/views/searchUserList/SearchUserList.test.js
```diff
@@ -535,7 +611,7 @@ describe('SearchUserList', () => {
       expect(wrapper.find('ReactTable').prop('columns')[1].id).toBe('enabled')
       expect(wrapper.find('ReactTable').prop('columns')[2].id).toBe('last_login_date_time')
       expect(wrapper.find('ReactTable').prop('columns')[3].accessor).toBe('racfid')
-      expect(component.find('ReactTable').prop('sorted')).toEqual([{ desc: 'value', id: 'last_name' }])
+      expect(component.find('ReactTable').prop('sorted')).toEqual([])
```
Nasir2712: Any reason to remove this?

app/javascript/views/searchUserList/SearchUserList.jsx
```diff
@@ -69,9 +75,10 @@ class SearchUserList extends PureComponent {
   }
 
   isDisabledSearchBtn = () => {
-    const { lastName, query } = this.props
+    const { lastName, firstName, query } = this.props
     const lastNameSearch = query.find(({ field }) => field === 'last_name')
-    return lastNameSearch && lastNameSearch.value === lastName
+    const firstNameSearch = query.find(({ field }) => field === 'first_name')
+    return lastNameSearch && lastNameSearch.value === lastName && (firstNameSearch && firstNameSearch.value === firstName)
```
Nasir2712: so assuming if more fields are added this check condition will get complicated?

kelfink: Yes.  And we may be in flux on future stories about what this expression should be.  For now this just extends the same logic as before.

app/javascript/views/searchUserList/SearchUserList.test.js
kelfink: Well it was failing, of course.  But I removed it from the setup because it doesn't represent the actual page.  we don't sort the react table, we sort the results in ES before populating this.

kelfink: Before, the sort was, I think, consistent with the ES result set being sorted by last name.  We are sorting now by a _score value from ES, so sorting here by last name would break that.

app/javascript/views/searchUserList/SearchUserList.jsx
```diff
@@ -154,6 +176,17 @@ class SearchUserList extends PureComponent {
 
   handleOnInput = (name, value) => this.props.actions.handleSearchChange(name, value)
 
+  handleEmailSearch = (name, value) => {
+    this.props.actions.handleSearchChange(name, value)
+    this.validateEmailField(value)
+  }
+
+  validateEmailField = value => {
```
RamuKammagani-osi: can you add test cases for this function??

kelfink: @Sushmareddy-OSI will handle this, I'm sure

RamuKammagani-osi:
retest this please

Sushmareddy-OSI:
retest this please

PR #390, opened 2019-04-17
FIT-558: Serve react-wood-duck fonts locally

edgibbs:
- FIT-558

-  To fix a security finding


edgibbs:
retest this please

docker/web/Dockerfile
```diff
@@ -9,10 +9,10 @@ RUN bundle install --jobs 20 --retry 5
 COPY package.json yarn.lock /app/
 RUN yarn install --production=false --non-interactive --frozen-lockfile
 COPY . /app
-RUN NODE_ENV=production RAILS_ENV=production bundle exec rails assets:precompile
+RUN NODE_ENV=production RAILS_ENV=production bundle exec rails assets:precompile && bundle exec bin/hack-to-copy-fonts.sh
```
kelfink: does this need to be a bundle exec ??
```suggestion
RUN NODE_ENV=production RAILS_ENV=production bundle exec rails assets:precompile && bin/hack-to-copy-fonts.sh
```

NagaPradeepGajula-osi: +1

edgibbs: It's safer

edgibbs: It guarantees it only runs with the gems in the Gemfile in case there are some globally installed gems, just considered good practice.

PR #392, opened 2019-04-18
Ignore any subqueries which don't specify a value

kelfink:
-  broken queries from latest changes

-  Some change recently is sending query parts for users with no 'value' node in the document.  We shouldn't do it, but we should be able to recover.  This avoids the failure.


kelfink:
retest this please

PR #394, opened 2019-04-19
Search by office fix

Sushmareddy-OSI:
-N/A

- Search by email / firstName / racfid is blocking due to the query passed in handleOfficeChange method. 
- Input field throws a warning `Input is changing uncontrolled input of type text to be controlled` since the value of fields are undefined in initial state.
- Search by case-insensitive racfid


app/javascript/containers/searchUserListContainer.js
```diff
@@ -41,10 +41,10 @@ function mapStateToProps(state) {
     error: searchUserList.error,
     officesList: officesList(state),
     inputData: searchUserList.inputData,
-    firstName: searchUserList.inputData.firstName,
-    email: searchUserList.inputData.email,
-    CWSLogin: searchUserList.inputData.CWSLogin,
-    lastName: searchUserList.inputData.lastName,
+    firstName: searchUserList.inputData.firstName ? searchUserList.inputData.firstName : '',
```
kelfink: doesn't safeGet do this for us?  import safeGet from 'lodash.get' then...
```suggestion
   firstName: safeGet(searchUserList, 'inputData.firstName', '')
```

Sushmareddy-OSI: yeah! good call out, isn't better to use the safeGet in the selector instead here in container?

kelfink: Sure.  That sounds even better.

Sushmareddy-OSI:
retest this please

lib/elastic/user_query_builder.rb
```diff
@@ -54,7 +54,8 @@ def self.query_leaves(query)
         { conjunction: 'AND', query: { match: { 'email.keyword': value } } } unless value.empty?
       end,
       racfid: lambda do |value|
-        { conjunction: 'AND', query: { match: { 'racfid.keyword': value } } } unless value.empty?
+        { conjunction: 'AND', query: { match: { 'racfid.keyword': value.upcase } } } unless value
```
kelfink: do we need the same logic on email.keyword, to upcase it ?

Sushmareddy-OSI: No not for the email.

PR #395, opened 2019-04-19
Cog 1339 fixing rails queries for users

kelfink:
-  COG-1339

- Fix a problem with detecting empty subquery values (not processed in Rails) versus value of false, which should be processed

### DIFF
--- a/spec/lib/elastic/user_query_builder_spec.rb
+++ b/spec/lib/elastic/user_query_builder_spec.rb
@@ -11,6 +11,29 @@ describe Elastic::UserQueryBuilder do
   end
 
   describe '.get_search' do
+    it 'processes enabled = false correctly' do
+      expected_output = {
+        query: {
+          bool: {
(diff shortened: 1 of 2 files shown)


spec/lib/elastic/user_query_builder_spec.rb
```diff
@@ -11,6 +11,28 @@
...
+              } }
+            ]
+          }
+        },
+        from: 0,
+        size: 10,
+        sort: sort_score_and_name
+      }
+      input_query = {
+        "query": [{ "field": 'enabled', "value": false }], "sort": [], "size": 10, "from": 0
```
RamuKammagani-osi: i guess we should check here without value in input_query?? just a question

kelfink: The block on line 74 "'processes the missing value parameters safely'". I think, tests this sufficiently.

PR #400, opened 2019-04-23
Cog 1413 perform search action

Sushmareddy-OSI:
[COG-1413](https://osi-cwds.atlassian.net/browse/COG-1413)

- perform search with existing search criteria on page reload


app/javascript/views/searchUserList/SearchUserList.test.js
```diff
@@ -742,46 +742,37 @@ describe('SearchUserList', () => {
     it('fetches the roles list', () => {
       expect(mockFetchRolesActions).toHaveBeenCalledWith()
     })
+
+    it('performes search with query', () => {
```
kelfink:
```suggestion
    it('performs search with query', () => {
```
:-)

app/javascript/views/searchUserList/SearchUserList.jsx
```diff
@@ -216,7 +224,7 @@ class SearchUserList extends PureComponent {
           )}
           <br />
           <div>
-            {this.props.searchedForUsers ? (
+            {!this.isDisabledAddUsrBtn() ? (
```
kelfink: I think now that this method is misnamed.  It really should be something like "criteriaSupplied".  Also, it seems to me the logic is changing here, right?  This logic controls the following div displaying if and only if any criteria have been displayed, and not whether a search has actually been performed.  Is that the intention here?

Sushmareddy-OSI: The actual logic is to display the table when search is performed, but it should not display on initial search. we can't really make it true and false in one same action(setSearch). so I am capturing the boolean value of `createUserButton` to display the table. Since create user button is enabled only on search(but not on initial search)

sorych: I agree with Kevin that this method maybe needs to be renamed as the same logic is leveraged for both button disabling and table rendering

PR #401, opened 2019-04-23
search with case-insensitive email value

Sushmareddy-OSI:
[COG-1342])

- Search with case insensitive email


Sushmareddy-OSI:
retest this please

lib/elastic/user_query_builder.rb
```diff
@@ -55,7 +55,8 @@ def self.query_leaves(query)
         { conjunction: 'AND', query: { terms: { 'office_id.keyword': value } } } unless value.empty?
       end,
       email: lambda do |value|
-        { conjunction: 'AND', query: { match: { 'email.keyword': value } } } unless value.empty?
+        { conjunction: 'AND', query: { match: { 'email.keyword': value.downcase } } } unless value
```
kelfink: interesting need for downcast here, yet upcase in the other field?  Could they both be upcase ?  Just curious.

kelfink: If they *could* both be the same casing, maybe we should be doing that for all searched fields...

Sushmareddy-OSI: firstName and LastName are already case-insenstive, thats how the dora is accepting..

PR #403, opened 2019-04-24
COG-1425 Dashboard Tiles Count Query Implementation

Nasir2712:
- https://osi-cwds.atlassian.net/browse/COG-1425

- A new dora endpoint is created to just return the count of users. This PR implements just the rails portion of that implementation.


Nasir2712:
jenkins, retest this please

Nasir2712:
jenkins, retest this please

lib/users/user_repository.rb
```diff
@@ -81,6 +81,12 @@ def self.search(query, auth_header)
       response.body
     end
 
+    def self.count(query, auth_header)
```
kelfink: I haven't researched the performance benefits of the newer Dora endpoint, but, while this count code looks good, I think it's worth noting that  the functionality of both this PR and the one which created the Dora _count endpoint could instead have been implemented by passing a size=0. parameter to the original search query.   "https://www.elastic.co/guide/en/elasticsearch/reference/current/returning-only-agg-results.html

Nasir2712: I mean still, research needs to be done on how to combine two terms and returning count.

kelfink: Agreed.  after our discussion, I think a good follow-up story would be to optimize this implementation by altering the index with 'copy_to' on the fields we want to aggregate ("group by") and perform one query against the index.

PR #415, opened 2019-04-30
Cog 1369 fix fuzzy seaerch

kelfink:
 COG-1369

- Fix the logic on fuzzy search / exact search splitting.  Considers exact match IFF all specified (non-blank) query params match the fields in the returned results exactly, but case-insensitive


RamuKammagani-osi:
@kelfink Shouldn't this be fix exactMatch?? COG-1369 is all about exact Match!

kelfink:
retest this please

PR #416, opened 2019-05-01
Protect against null / undefined values in user records

kelfink:
- COG-1369

- Protect against comparison of missing values like racism fid which might be missing or undefined in a user record, but need to be compared against query params


mgrmax:
retest this please

PR #426, opened 2019-05-11
Cog 1436 office admin user acceptance test

kelfink:
- COG-1436 acceptance test
- Some refactoring of helper method names
- Splitting of tests into office and county for different users
- Improvements to detecting when verification code entry is completed


spec/acceptance/county/user_login_spec.rb
```diff
@@ -92,8 +92,7 @@
     expect(page)
       .to have_text('Error. Incorrect code. You have 2 attempts remaining.')
     verify_account
-    go_manage_users
```
RamuKammagani-osi: We still need go_manage_users to land on Search Page from dashboard! isn't it??

kelfink: ah, maybe!  This might explain the PR failures.  Let me try....

kelfink:
Made changes to the API requests, so it grabs them relative to the /cap which is not always in place in the various environments.

spec/acceptance/county/user_login_spec.rb
kelfink: Failure in the PR ended up being URL confusion in the tests.  was directing to the wrong place.  See *green* builds now.

kelfink: PR issue was fixed by URL changes in later commits.

PR #429, opened 2019-05-13
Cog 1340 sort tuning

kelfink:
- COG-1340

- Sorting improvement.  Fuzzy and prefix name searches boosted the same as phonetic.


kelfink:
retest this please

PR #433, opened 2019-05-15
Boost all name searches equally

kelfink:
- COG-1340

- Tuning to boost all matching types for first and last name.


sorych:
you have a boost value set in multiple places and you have already changed it from 3 to 5 - is not it a good candidate to introduce a variable?

PR #434, opened 2019-05-15
COG-1436 Fix office selection when no value

RamuKammagani-osi:
- https://osi-cwds.atlassian.net/browse/COG-1436

- Office value preselects even after removing when visiting to detail and coming back to search page.

### DIFF
--- a/app/javascript/reducers/searchUserListReducers.js
+++ b/app/javascript/reducers/searchUserListReducers.js
@@ -99,14 +99,6 @@ function searchUserListReducer(state = initialValue, { type, payload, error, met
     case actionTypes.USER_LIST_CLEAR_SEARCH:
       return {
         ...state,
-        // inputData: {
-        //   ...state.inputData,
-        //   officeNames: getAdminOfficeIDs(state.adminAccountDetails),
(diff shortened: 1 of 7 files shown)


app/javascript/reducers/searchUserListReducers.js
```diff
@@ -120,10 +113,7 @@ function searchUserListReducer(state = initialValue, { type, payload, error, met
         fetching: false,
         inputData: {
           ...state.inputData,
-          officeNames:
-            state.inputData.officeNames === undefined || state.inputData.officeNames.length === 0
-              ? getAdminOfficeIDs(payload)
-              : state.inputData.officeNames,
+          officeNames: state.inputData.officeNames === undefined ? getAdminOfficeIDs(payload) : state.inputData.officeNames,
```
kelfink: Looks good.  simpler than before.   Do we not need to change unit testing for this no logic?  Seems like you would have had a test before for length == 0, and now it should fail.

RamuKammagani-osi: We don't have a unit test for .length, anyway we removed it!

PR #437, opened 2019-05-17
User detail page to load logged-in-user's info. Fix county/office acceptance

kelfink:
- COG-1436

- further fixes to acceptance testing


spec/acceptance/office/search_user_list_page_spec.rb
```diff
@@ -30,11 +30,11 @@
...
     page_is_search
-    # page_has_user_list_headers
-    search_users(last_name: 'Manzano')
-    my_account = logged_in_account
+    current_account = logged_in_user_info
 
-    expect(selected_offices[0]).to eq office_name_map[my_account[:admin_office_ids][0]]
+    search_users(last_name: 'M')
+
+    expect(selected_offices[0]).to eq current_account[:office_name]
```
kelfink: This will fail when running the 'office' suite until COG-1436 bug is fixed.

Sushmareddy-OSI: I don't think so, because the bug is not on the initial search.

kelfink: Ash but this isn't actually the initial search, due to the load of the account info, another search is already performed, and the CLEAR button was clicked

PR #439, opened 2019-05-17
COG-1453 Column Sorting available in the Filter User Lists

AlexanderSerbin:
https://osi-cwds.atlassian.net/browse/COG-1453

Added custom sorting for Last Login column


app/javascript/views/groupUsersList/GroupUsersList.jsx
```diff
@@ -23,7 +23,15 @@ class GroupUsersList extends PureComponent {
           columns={[
             { Header: 'Full Name', id: 'last_name', accessor: toFullName, minWidth: 200 },
             { Header: 'Status', id: 'enabled', accessor: accountStatusFormat, minWidth: 60 },
-            { Header: 'Last Login', id: 'last_login_date_time', minWidth: 150, accessor: lastLoginDate },
+            {
+              Header: 'Last Login',
```
Nasir2712: is there a test case already written for this?

app/javascript/views/groupUsersList/GroupUsersList.jsx
```diff
@@ -23,7 +23,15 @@ class GroupUsersList extends PureComponent {
...
+              id: 'last_login_date_time',
+              minWidth: 150,
+              accessor: lastLoginDate,
+              sortMethod: (a, b) => {
+                return new Date(b) - new Date(a)
```
kelfink: The sortMethod in react-table is not well specified, but all the examples I have seen return either 1 or -1.  Is this working as intended?

kelfink: Best documentation of the method is here: https://www.npmjs.com/package/react-table
So I'd suggest replacing the return with :
return (new Date(b) > new Date(a)) ? 1 : -1
```suggestion
                return new Date(b) - new Date(a)
```

app/javascript/views/groupUsersList/GroupUsersList.jsx
AlexanderSerbin: Added tests for date time comparator

app/javascript/views/groupUsersList/GroupUsersList.jsx
AlexanderSerbin: Done. Changed comparator function to "return new Date(a) > new Date(b) ? 1 : -1"

PR #441, opened 2019-05-20
Cog 1328 enable reset button on detail page

Sushmareddy-OSI:
[COG-1328](https://osi-cwds.atlassian.net/browse/COG-1328)

-RESET button in user profile page is not enabled after user makes changes in email or phone number fields.


app/javascript/selectors/detailSelector.test.js
```diff
@@ -531,40 +531,34 @@ describe('selectors', () => {
...
-      expect(disableActionButton(state)).toEqual(true)
+      const state = getState({ phoneNumber: '333-444' })
+      expect(disableSaveButton(state)).toEqual(true)
     })
 
     it('return false when email & phone number is valid', () => {
       const state = getState({
         phoneNumber: '1234567891',
-        email: 'hello@gmail.com',
+        email: 'hel@gmail.com',
```
kelfink: why this change?  Not objecting to it, I just don't know if it's relevant.  Does changing the email from hello (possibly in the before state). to this  change the logic?

Sushmareddy-OSI: Uhh No..., no logic this is accidental.

PR #450, opened 2019-06-04
style fix for scrollbar

Sushmareddy-OSI:
[COG-1171](https://osi-cwds.atlassian.net/browse/COG-1171)

- ScrollBar missing for ChangeLog Table in UserList and UserProfile Page


app/javascript/views/userDetail/CWSPrivileges.jsx
```diff
@@ -39,6 +39,7 @@ const CWSPrivileges = ({ CWSPrivileges }) => (
           defaultPageSize={100}
           style={{
             maxHeight: '500px',
+            overflowY: 'scroll',
```
kelfink: would 'auto' be sufficient here, or was it already auto and wasn't working?

app/javascript/views/userDetail/ChangeLog.jsx
```diff
@@ -143,6 +143,7 @@ const ChangeLog = ({
           defaultPageSize={100}
           style={{
             maxHeight: viewHeightSize(isListView),
+            overflowY: 'scroll',
```
kelfink: same question as above

PR #454, opened 2019-06-08
Only rollback deploy on smoketest failures

kelfink:
- FIT-580

-  fix rolls back too aggressively, undoing rollouts when any acceptance test fails.  Only smoke test failures should prevent keeping the releases in Integration.


kelfink:
retest this please

PR #456, opened 2019-06-10
Fix: FIT-580. removed acceptance tests from the rollback block.

LotharII:
-https://osi-cwds.atlassian.net/browse/FIT-580

- removed acceptance tests from the rollback block.

### DIFF
--- a/JenkinsDeploy
+++ b/JenkinsDeploy
@@ -33,10 +33,13 @@ def deployWithSmoke(environment){
       deployToStage(environment, env.version)
       updateManifestStage(environment, env.version)
       if (environment == 'integration')  {
-          smokeTest(environment, env.version)
-          integrationAcceptenceTest(environment, env.version)
+        smokeTest(environment, env.version)
       }
     }
+    if (environment == 'integration')  {
+      integrationAcceptenceTest(environment, env.version)
+    }
+
   }
 }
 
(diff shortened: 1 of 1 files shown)


JenkinsDeploy
```diff
@@ -33,10 +33,13 @@ def deployWithSmoke(environment){
       deployToStage(environment, env.version)
       updateManifestStage(environment, env.version)
       if (environment == 'integration')  {
-          smokeTest(environment, env.version)
-          integrationAcceptenceTest(environment, env.version)
+        smokeTest(environment, env.version)
       }
     }
+    if (environment == 'integration')  {
```
kelfink: Ah.  This is a good addition.  wrapping it in the test for integration is probably necessary.  Unfortunately I just merged my simpler solution. If you have time to resolve the (my) conflict I'd like to add this change as well.

LotharII: Sure, thanks.

LotharII: I will just create a new PR with wrapping acceptance tests call.

kelfink: Well, up to you.  I resolved the conflict.  If that looks good, go ahead and merge.

kelfink: Oh my merge isn't good enough.  I put the full integration test run outside the node.  It's no good.  I'll try another PR.

PR #468, opened 2019-06-25
Include fuzzy counts in the match count

kelfink:
- COG-955.

-  A remaining test fails because the exact count and fuzzy count are correctly split, but the test expected to find a fuzzy match.  So, fixed that by adding the fuzzy match to the total count it expects.


kelfink:
retest this please

PR #486, opened 2019-11-08
CMO 285 User Profile Performance

TylerOSI:
- [CMO 285](https://osi-cwds.atlassian.net/browse/CMO-285)

- decrease page size to 20 for Change Log
- Fix header to display 20 or less


TylerOSI:
Jenkins test this please

PR #487, opened 2019-11-12
revert yarn.lock to what it was before the performance changes

TylerOSI:
- [CMO-285](https://osi-cwds.atlassian.net/browse/CMO-285)

- Reverting the lock file to what it was before the performance changes to fix styling issues
- There were also some eslint changes that needed to be reverted and a failing test that needed to be fixed


TylerOSI:
Jenkins test this please

PR #500, opened 2020-03-04
CMO-424: CAP hangs in IE

david-smith-osi:
- CMO-424

- CAP hangs in IE, when super users display all active users.


david-smith-osi:
retest this please

